PR #51, opened 2019-02-12
Refactor DashR callbacks for increased parity with Dash for Python API 

rpkyle:
The proposed changes include
- ~the `callback` method for Dash now uses the `output, inputs, state, func` ordering for parameters, while previously `output` was last; the change reflects the current Dash for Python callback signature~
- the `callback` method for DashR now has a function signature that is closer to the Dash for Python API, with the caveat that `inputs` and `state` are unified into a single `params` argument
- the addition of a `func` argument, to pass the callback handler; R does not natively offer a Python-like decorator pattern, which is the current approach for specifying this function in Dash for Python
- basic validation of callbacks using `assert_valid_callbacks`
- updating `dash-renderer` to 18.0

The current batch of examples will have to be updated to reflect the changes. I'm continuing to work on the code, but callbacks appear to function properly now.


R/dash.R
```diff
@@ -214,19 +217,14 @@ Dash <- R6::R6Class(
...
           list(
-            output = list(id = y[[1]], property = y[[2]]),
-            inputs = setNames(callback_inputs(x), NULL),
-            state = setNames(callback_states(x), NULL),
-            # Chris mentioned that events might/should be deprecated
-            events = setNames(callback_events(x), NULL)
+            output=callback_signature$output,
+            inputs=callback_signature$inputs,
+            state=callback_signature$state,
+            key=paste(callback_signature$output$id, callback_signature$output$property, sep='.')
```
chriddyp: Is `key` a special keyword in R's lists? I'm assuming so, since I don't see this referenced anywhere else in the code.

R/dash.R
```diff
@@ -374,55 +344,24 @@ Dash <- R6::R6Class(
     # ------------------------------------------------------------------------
     # callback registration
     # ------------------------------------------------------------------------
-    callback = function(func = NULL, output = NULL) {
+    callback = function(output = NULL, inputs = list(), state = list(), user_function) {
```
chriddyp: I think we should rename `user_function` into something more human-readable / conventional. The natural name would be, well, `callback`, but that feels redundant. Maybe `callback_func` or just `func`?

R/dash.R
```diff
@@ -374,55 +344,24 @@ Dash <- R6::R6Class(
...
+      private$layout_render()
```
chriddyp: Why is this called?

R/dash.R
```diff
@@ -374,55 +344,24 @@ Dash <- R6::R6Class(
...
-        stop("The layout must be set before defining any callbacks", call. = FALSE)
-      }
+      assert_valid_callbacks(output, inputs, state, user_function)
 
-      # -----------------------------------------------------------------------
-      # verify that output/input/state IDs provided exists in the layout
-      # -----------------------------------------------------------------------
-      callbackInputs <- callback_inputs(func)
-      callbackStates <- callback_states(func)
+      # set class attributes for compatibility with original DashR code
```
chriddyp: Can we delete this in the future? I don't think we should worry about backwards compatability until we officially release.

R/utils.R
```diff
@@ -7,10 +7,14 @@ is.fire <- function(x) inherits(x, "Fire")
 
 # dependencies
 is.dependency <- function(x) inherits(x, "dash_dependency")
-is.output <- function(x) is.dependency(x) && inherits(x, "output")
-is.input <- function(x) is.dependency(x) && inherits(x, "input")
-is.state <- function(x) is.dependency(x) && inherits(x, "state")
-is.event <- function(x) is.dependency(x) && inherits(x, "event")
+#is.output <- function(x) is.dependency(x) && inherits(x, "output")
```
chriddyp: 🔪  - in general, let's delete code instead of commenting it out. It'll always be available in git if we want to refer to it later

README.md
```diff
@@ -49,11 +49,10 @@ app$layout_set(
   htmlDiv(id = "outputID")
 )
 
-app$callback(
-  function(x = input("inputID"), y = input("inputID", "type")) {
+app$callback(output=list("outputID"), inputs=list(list("inputID", "type")), user_function=
```
chriddyp: Is `user_function=` needed? Or can a user just pass in `function(value) {...}`?

README.md
```diff
@@ -49,11 +49,10 @@ app$layout_set(
...
+  function(x, y) {
```
chriddyp: This should just be a single input argument now, right?

README.md
chriddyp: Also, should this be `output=list("outputID", "children")`?

chriddyp: Also, for the pedagogical example, I think we should do `inputs=list(list("inputID", "value"))`, since that's more useful than `type` (`type` doesn't change during runtime, but `value` changes when the user types into the input)

README.md
```diff
@@ -110,15 +109,13 @@ app$callback(
...
 )
 
-app$callback(
-  function(x = input("graphTitle"), y = input("graphTitle", "type")) {
+app$callback(output=list("outputID"), inputs=list(list(id="graphTitle", property="type")), user_function=
+  function(x, y) {
     sprintf("You've entered: '%s' into a '%s' input control", x, y)
-  },
-  output("outputID")
+  }
```
chriddyp: similarly down here

R/dash.R
rpkyle: This was present in the original code, but I can remove it. The key is also referenced here:

https://github.com/plotly/dashR/blob/b05eaed4ed77ce625ce7e794e2bc80aa622cd640/R/dependencies.R#L55

R/dash.R
rpkyle: Agreed, I think `func` would be preferable instead.

R/dash.R
rpkyle: If you comment out that line, the app does not finish rendering. This line was originally present in Carson's code (see below); I deleted it after you and I spoke last week, but reinserted it after I noticed the page never finished loading.

https://github.com/plotly/dashR/blob/41e53e3cc5bdfcbfb6e94373ac28d4939b001457/R/dash.R#L384

R/dash.R
rpkyle: Yes; we don't actually need the class attributes, but I can understand why Carson used them.

R/utils.R
rpkyle: Right, sorry about this; I had commented out these lines and others to test, will :hocho:.

README.md
rpkyle: The way the `callback` method is currently defined, `user_function=` must be specified, or the following error message is returned:

``Error in assert_valid_callbacks(output, inputs, state, user_function) : 
  argument "user_function" is missing, with no default``

rpkyle: > Also, should this be `output=list("outputID", "children")`?

While the app seems to run without the modification, I think it makes sense to change this.

rpkyle:
> Looking good! Just a few small comments & questions

Thanks for taking the time to review. I'll make these initial edits and await Nadia and Alex's suggestions.

README.md
alexcjohnson: I actually kind of like having to use a special function/constructor for the references here. So rather than `output=list("outputID", "children")` you'd just have `output("outputID", "children")`. Seems to me this has a few benefits:
- it'll immediately tell you if you forgot the `"children"` - and presumably the IDE can tell you as soon as you type `output(` that you need an id and a property. That said, I notice in the original code `"children"` is the default property (and `"value"` for input/state). I don't think that's a good idea, we should force users to be explicit about the property being referenced. (also there's still an `event` in that file)
https://github.com/plotly/dashR/blob/5f89e8034c793e2ae37266b58251887542280667/R/dependencies.R#L16
- More of the syntax conveys meaning, no extra `list`. You'd still have `list` for the inputs, but only the outer one: `list(input("inputID", "value"))`
- This would differ from the python side, but what if we combined inputs and state into a single list like `list(input("inputID", "value), state("input2ID", "value"))? We would need to check and throw an error if there isn't at least one `input` in the list, and we could discuss whether all the inputs need to be before all the states or if we can accept arbitrary order, but the advantage is then all arguments are required - you just have `output`, `inputs_and_state`, and `func` - so the user doesn't need to name any of them.

R/utils.R
```diff
@@ -7,10 +7,14 @@ is.fire <- function(x) inherits(x, "Fire")
...
+is.output <- function(x) return(x)
+#is.input <- function(x) is.dependency(x) && inherits(x, "input")
+is.input <- function(x) return(x)
+#is.state <- function(x) is.dependency(x) && inherits(x, "state")
+is.state <- function(x) return(x)
+#is.event <- function(x) is.dependency(x) && inherits(x, "event")
+is.event <- function(x) return(x)
```
alexcjohnson: 🔪 `event`

R/dash.R
```diff
@@ -513,21 +452,21 @@ Dash <- R6::R6Class(
...
         } else {
-          return(NULL)  
+          return(NULL)
         }
       })
-      
+
       deps_layout <- unlist(deps_layout, recursive=FALSE)
 
       # if core components are used, but no coreGraph() exists,
```
alexcjohnson: Not part of this PR, but I don't think we can filter out plotly.js this way - you never know what elements will get added by a callback.

R/dash.R
alexcjohnson: If we can't figure this out in this PR please make an issue for it. For a complicated app this would be a major drain, re-rendering the layout for every callback you attach. It'll only happen on app initialization, but if you have a slow layout and lots of callbacks it would be painful.

dash-py does renders the layout once in `layout.setter` - perhaps we can remove it here and add it to `layout_set`?

rpkyle: As discussed earlier today, will attempt to relocate the `layout_render()` call to the `layout_set()` method.

README.md
rpkyle: Thanks for these suggestions; I like this approach as well. The `state` parameter should still be optional, correct? If true, I think we'd set no default for the `input` parameter, but use `NULL` as the default for `state`.

alexcjohnson: I thought what we settled on at the meeting was to combine input and state into one parameter (that I was calling `inputs_and_state` but could just be `params` or something). Then we set no defaults for any of the parameters, as they're all required.

R/dash.R
rpkyle: As we discussed, this option is only supposed to be used in one case, when the app developer wishes to sidestep including a Graph component in the initial layout, but does want one following a callback:

https://github.com/plotly/dashR/blob/5f89e8034c793e2ae37266b58251887542280667/R/dash.R#L49

R/utils.R
rpkyle: :bomb: `event`

README.md
rpkyle: Yes, as we discussed, this is a great suggestion. Will use class attributes to differentiate between inputs and state, and set them all in one shot.

R/utils.R
```diff
@@ -261,35 +256,38 @@ filter_null <- function(x) {
   x[!vapply(x, is.null, logical(1))]
 }
 
-assert_valid_callbacks <- function(output, inputs, state, user_function) {
+assert_valid_callbacks <- function(output, params, func) {
+  inputs <- params[vapply(params, function(x) 'input' %in% attr(x, "class"), FUN.VALUE=logical(1))]
+  state <- params[vapply(params, function(x) 'state' %in% attr(x, "class"), FUN.VALUE=logical(1))]
```
alexcjohnson: This is filtering `params` for either `input` or `state` objects, I gather. Seems to me there are two more things we need to test:
- there's nothing in `params` that's neither `input` NOR `state`
- the first `state` (if any) comes after the last `input`

rpkyle: Yes, that's correct; I'll add a check to verify that `params` only contains elements with `class` of `input` or `state`.

Can you explain why `state` must follow `input`? I understand that this is the convention, but one of the side effects of your suggested interface change is that the order is now unimportant, since we're assigning attributes via the `input` and `state` wrapper functions.

rpkyle: The `vapply` line here is essentially a filter, which returns `TRUE` if the list element's `class` attributes include `input` or `state`. One could use `sapply` instead, but that function is not type safe. `vapply` requires you to provide `FUN.VALUE`, which addresses that issue. 

In some instances, `vapply` is also faster than `sapply`.

rpkyle: > * there's nothing in `params` that's neither `input` NOR `state`

fixed in 61feba58d2f60caf45b1468cfdd5fb28cca6a1e0

alexcjohnson: > Can you explain why `state` must follow `input`?

Correct me if my lack of R experience is mangling this...

It would be possible to relax this constraint, but as it stands, we split it up into two lists internally, and provide these separately to `dash-renderer`. When `dash-renderer` then makes the request associated with this callback, it sends us two lists of values. We use these to call the callback:
```r
output_value <- do.call(callback, args = as.list(request$body$inputs, request$body$state))
```
So if I'm understanding it correctly, we'll pass these to the callback with all the inputs first, then the state values. If that doesn't match the order they had in `params`, we'll have to at that point reshuffle these values into the `params` order. So in the end, assuming I have that right, I guess it's your call: would it be easier to enforce the ordering of inputs then state, or reshuffle elements here? I'd be fine with it either way; the ordered version is closer to dash-py, the reshuffled is more flexible.

README.md
rpkyle: Each `id`/`property` pair is a single call to the `input` function.

R/utils.R
rpkyle: > * the first `state` (if any) comes after the last `input`

fixed in 703215088f1170e1922b0a14ef22c63eee2e05d5

rpkyle: It makes sense to me to ensure `input`/`state` ordering.

The current solution uses R's run-length encoding function to handle the case when `input` or `state` is preceded and followed  by `state` or `input`, respectively. I imagine there are much better approaches, but this does the job for now.

alexcjohnson: Thanks 🎉
`rle` looks fine to me. I wouldn't worry about performance here, it's not a hot path.

README.md
alexcjohnson: We still want to change this, right?
```R
app$callback(output("outputID", "children"), list(input("inputID", "value")), function(x) {
```

README.md
alexcjohnson: As shown above (or in my understanding of how we want it modified https://github.com/plotly/dashR/pull/51/files/5f89e8034c793e2ae37266b58251887542280667#r258077217) there's only one id/prop pair now, so this function should only take one arg, right?

README.md
alexcjohnson: Oh I see, the original callbacks here actually did have two inputs. Perhaps we can use that in at least one of these cases to show how to use `state`:
```R
app$callback(output("outputID", "children"),
  list(input("graphTitle", "value"), state("graphTitle", "type")),
  function(x, y) {
    sprintf("You've entered: '%s' into a '%s' input control", x, y)
  }
)
```

rpkyle:
@alexcjohnson @chriddyp @TahiriNadia 

I've updated the examples in `README.md`, in case you're able to take a quick look. Hopefully we're nearing a point where we can :dancer: this one.

README.md
rpkyle: As we discussed, the `README.md` is now updated. I'll go ahead and trim out the declarations of `id` and `property` since they're implicit if unspecified.

README.md
rpkyle: I can simplify the example so that the formals don't include the argument names.

R/dash.R
alexcjohnson: These `key=` lines are still present 🔪

R/dash.R
alexcjohnson: Ha OK, I guess we're going to get the removal of this option from master when this PR is merged... https://github.com/plotly/dashR/commit/7663d584926ca0cb49797f7c122da9a30af5910b

R/dash.R
```diff
@@ -328,6 +301,7 @@ Dash <- R6::R6Class(
     },
     layout_set = function(...) {
       private$layout <- if (is.function(..1)) ..1 else list(...)
+      lay <- private$layout_render()
```
chriddyp: This line confuses me a bit, although it might just be my misunderstanding of R. Where is `lay` being used? 
- Is this setting it to the class? 
- Does this also implicitly return `lay`?

chriddyp: Also, why are we calling `layout_render` here? Could we leave a comment?

R/dash.R
rpkyle: fixed in 7e51f91b2dc11155043c5957cf527d568c8f66b0

rpkyle: fixed in 313dff1a9c99826fa05bea285fec73ac49e59169

R/dash.R
rpkyle: Currently, if you don't assign `layout_render()` to an object, it returns the layout, which will display in the console.

There are a couple ways I could handle this -- I can wrap the return in `invisible`, so that it silently returns the layout, which can be assigned to a new object. Or, I can add a `silent` option which is `TRUE` by default and let the app developer decide.

rpkyle: > Also, why are we calling `layout_render` here? Could we leave a comment?
I mentioned this last week, but I think it got lost in the shuffle. If we don't render the layout when it's set, I don't believe it will get rendered until a `GET` request unless `layout_get(render=TRUE)` is called.

chriddyp: > it returns the layout, which will display in the console.

Because if you didn't return the object but still needed to call `private$layout_render()`, the code would look like:
```
layout_set = function(...) {
       private$layout <- if (is.function(..1)) ..1 else list(...)
       private$layout_render()
}
```

which would actually return `private$layout_render()` (since it's the last line of the function).

But if we didn't need to call `private$layout_render()`, then the function would just be 
```
layout_set = function(...) {
       private$layout <- if (is.function(..1)) ..1 else list(...)
}
```
which wouldn't return anything and be OK.

Am I understanding this correctly?

In Python, there is a convention to call "unused / do not care" arguments `_` (https://hackernoon.com/understanding-the-underscore-of-python-309d1a029edc). Not sure if you can assign variables to `_` in R or if R has that convention, but this codeblock might be easier for other folks to understand if it was written:
```
layout_set = function(...) {
       private$layout <- if (is.function(..1)) ..1 else list(...)
       # call layout_render() in order to  [...] but don't return its response
       _ <- private$layout_render()
}
```

chriddyp: anyway, this isn't blocking, just trying to understand things a bit better!

rpkyle: Oh, no worries -- I appreciate your questions because they often lead me to a better understanding of the way Dash works.

> This line confuses me a bit, although it might just be my misunderstanding of R. Where is `lay` being used?
> 
> * Is this setting it to the class?
> * Does this also implicitly return `lay`?

The code (which I've since modified) does not set class attributes or implicitly return `lay` since this line isn't the last one in the function. I've since wrapped `layout_render()` inside a call to `invisible`, which I think has the same effect as assignment to `_` in Python. (I actually did not know about that, but it's a cool feature of which I should be aware.)

I've added a brief comment to explain what's happening here also.

alexcjohnson: Yep, would be nice to understand at some point why this call is necessary (for that matter, I don't quite get why we have a [nearly identical call](https://github.com/plotly/dash/blob/5b155f1e3690904a938040a84fdf2d4a20682e99/dash/dash.py#L297) on the Python side 😏) but thanks for making the intention clearer with `invisible`.

rpkyle: fixed in f9c9174045abfea1593db06f0c5f4ce77a4a1b2c

rpkyle: Missed a question or two here:

>which would actually return `private$layout_render()` (since it's the last line of the function).

Yes, that's right.

>But if we didn't need to call `private$layout_render()`, then the function would just be

Yep, got it -- this line basically says _if the first argument of `...` is a function, assign it to `layout`, otherwise wrap all the arguments into a list and assign that into `layout`.

Issue #27, opened 2018-03-20
Blacklist unsafe ports


cpsievert:
Similar to https://github.com/rstudio/shiny/commit/4fa2af72cc74bf72dc807e1d1709b13711a8a635

Perhaps, this could be handled on the fiery-level, @thomasp85?

PR #53, opened 2019-02-21
[WIP] Inspect environment variables and set routing prefixes, add catchall endpoint

### LINKED ISSUE #52: need a catchall endpoint
Currently, the app only responds when you visit http://localhost:8080. However, the app should respond to _any_ page, for example http://localhost:8080/some-other-page.

Webservers usually have dynamic / wildcard endpoints. We basically want to serve the HTML string for _all_ endpoints except the API endpoints (like `/_dash-layout`). That way, users can write multi-page apps.

rpkyle:
This PR will introduce a 'catchall' endpoint `/*`, which will be added as the final handler in the `routr` stack.

Additionally, DashR does not inspect local environment variables when initializing an app; `routes_pathname_prefix` and `requests_pathname_prefix` must be set manually. For greater parity with Dash for Python, the following route-related variables should be checked before setting route prefixes to `NULL`:
- ~`DASH_URL_BASE_PATHNAME`~
- `DASH_ROUTES_PATHNAME_PREFIX`
- `DASH_REQUESTS_PATHNAME_PREFIX`

Two additional helper functions are proposed in this PR:
- `get_package_mapping`: given path to dependency, package string from request URL, and list of dependencies, attempts to build a `package_map`. The mapping is then used to resolve local paths to dependencies within the R packages that provide them. A list is returned with the R package name providing the dependency, as well as a local path to the file.
- `get_mimetype`: resolve dependency MIME type based on file extension, currently supports `application/JavaScript`, `application/json`, `text/css`.

Lastly, previous versions of DashR did not include a `_dash-component-suites` handler. This PR also
- eliminates static handling of routes for Dash components served upon index loading using `ressource_route` (which resolves a duplicate route error for the React JS upon relaunching an existing DashR app in the current workspace) 
- related to item above, :hocho: `dashR::resolve_dependencies` and `register_dependencies` 
- constructs script tags manually, rather than with `htmltools`
- provides a `dash_suite` handler to serve dependency files, following the Dash convention
- provides support for inserting the DashR app name into dependency pathnames
- provides support for including a `Cache-Control` header, and setting the cache lifetime via `components_cache_max_age`
- more careful CSS handling
- treats dependencies bundled into DashR (i.e. `dash_renderer`, `react`, `react-dom`) as being part of the `dash_renderer` "package", to mirror behaviour in Dash for Python
- some basic sanity checking for dependencies: removing multiple slashes in pathnames, ensuring that source maps aren't being served until user interacts with debugger, checking for duplicate dependencies

This PR is a work in progress, aiming to implement similar behaviour as found in `_configs.py`:

https://github.com/plotly/dash/blob/f590166be2ca476c8970512fd9b23aaf26541ae8/dash/_configs.py#L8

...and in `dash.py`:

https://github.com/plotly/dash/blob/1249ffbd051bfb5fdbe439612cbec7fa8fff5ab5/dash/dash.py#L488

Closes #52.


chriddyp:
I think let's 🔪 `DASH_URL_BASE_PATHNAME` - We've only included it for backwards compatible reasons in dash.py. Instead, we can force users to be more explicit by setting both `requests` and `routes` prefixes

R/dash.R
```diff
@@ -268,7 +268,14 @@ Dash <- R6::R6Class(
         TRUE
       })
 
-
+      route$add_handler('get', '/*', function(request, response, keys, ...) {
```
chriddyp: Did you have to move this around so that it gets declared _after_ the other routes or does it work like this?

R/dash.R
```diff
@@ -370,8 +377,32 @@ Dash <- R6::R6Class(
     # private fields defined on initiation
     name = NULL,
     serve_locally = NULL,
-    routes_pathname_prefix = NULL,
-    requests_pathname_prefix = NULL,
+    routes_pathname_prefix = {
```
chriddyp: This syntax is interesting - is this like a function that gets called whenever this variable is accessed? Or is it just a function shorthand?

R/dash.R
chriddyp: Oh nvm, the original route wasn't removed: https://github.com/plotly/dashR/pull/53/files#diff-1881af3720f8f49fd3a44e51c9fb5a0dR180

Let's 🔪 the original route since `/*` should catch `/`. 

Also, this route needs to be prefixed by `routes_pathname_prefix`

R/dash.R
```diff
@@ -370,8 +377,32 @@ Dash <- R6::R6Class(
...
+      routes_prefix <- Sys.getenv("DASH_ROUTES_PATHNAME_PREFIX")
```
chriddyp: These variables are also defined in the constructor: 
```
routes_pathname_prefix = '/',
requests_pathname_prefix = '/',
```

The order of preference should be:
1. Did the user supply it in the constructor? If so, then use that.
2. If not, then check the environment
3. If not in the environment, then use the default.

In Python, for (1), the only way to check if a user supplied it while keeping a default argument was to set that default argument to `None`. That way, if it's not `None`, then we can assume that the user set it. (If the default was `/`, then we wouldn't know whether the user explicitly supplied that or if its just the default).

R/dash.R
rpkyle: Right, I noticed that when I glanced at the Python code but forgot to insert here.

R/dash.R
rpkyle: This is just some logic to set the initial value, it’s on the private side, so cannot be modified directly from the console.

R/dash.R
rpkyle: OK, perfect. This makes sense. I’ll just use `NULL` instead of `None`.

R/dash.R
rpkyle: I realized this isn't going to work the way we want, so I've :hammer: a bit, hopefully the refactored implementation of the routes/requests pathname prefixes is closer to the Python logic.

R/dash.R
rpkyle: fixed in 5fba10f16601ebe26b15f3b96334105b7eafff31

R/dash.R
rpkyle: fixed in 62d7dd9965e8beb42002031830c6c30c377fcb1f

rpkyle:
fixed in 5fba10f16601ebe26b15f3b96334105b7eafff31

rpkyle:
> I think let's 🔪 `DASH_URL_BASE_PATHNAME` - We've only included it for backwards compatible reasons in dash.py. Instead, we can force users to be more explicit by setting both `requests` and `routes` prefixes

fixed in 5fba10f16601ebe26b15f3b96334105b7eafff31

rpkyle:
@chriddyp @alexcjohnson I've updated the list of proposed changes, and tried to structure the commits in a (slightly) more thoughtful way. I look forward to your comments and suggestions, hopefully we're inching the :football: a bit further towards the deployment goal line with this one.

R/dash.R
```diff
@@ -112,16 +112,15 @@ Dash <- R6::R6Class(
                           server = fiery::Fire$new(),
                           static_folder = NULL,
                           serve_locally = TRUE,
-                          routes_pathname_prefix = '/',
-                          requests_pathname_prefix = '/',
-                          suppress_callback_exceptions = FALSE) {
+                          routes_pathname_prefix = NULL,
+                          requests_pathname_prefix = NULL,
+                          suppress_callback_exceptions = FALSE,
+                          components_cache_max_age = 2678400) {
```
chriddyp: 🔪 `cache_max_age` - As we discussed today, this will be handled via the query strings and the cache will be infinite.

R/dash.R
```diff
@@ -176,16 +175,16 @@ Dash <- R6::R6Class(
       # ------------------------------------------------------------------------
       route <- routr::Route$new()
 
-      dash_index <- routes_pathname_prefix
+      dash_index <- self$config$routes_pathname_prefix
       route$add_handler("get", dash_index, function(request, response, keys, ...) {
```
chriddyp: do we still need this given the catch all?

R/dash.R
```diff
@@ -257,17 +256,57 @@ Dash <- R6::R6Class(
...
+        #
+        # 1) extract the URL from the request
+        # 2) split the URL into chunks for each of the /*/ blocks
+        # 3) retrieve the script pathname by starting from the first block following
+        #    the "package" name (for the Dash deps, not the R package) name in the
+        #    URL, then capturing the rest of the pathname afterwards
+        # 4) concatenate the strings into a path, collapsing with /
+        # 5) pass the constructed path to get_package_mapping to retrieve R pkg name
+        url <- request$url
+        url_as_strings <- strsplit(gsub("http://|https://|www\\.", "", url), "/")[[1]]
```
chriddyp: It might be cleaner to just use the path pattern matching itself. If you write your URL like:
```
       dash_suite <- paste0(self$config$routes_pathname_prefix, "_dash-component-suites/:package_name/:filename")
```

Then I believe that:
```
dist_path = keys$filename
dep_pkg = keys$package_name
```
will be what you're looking for.

R/utils.R
```diff
@@ -327,4 +335,66 @@ valid_seq <- function(params) {
...
+}
+
+get_package_mapping <- function(script_path, url_package, dependencies) {
+  package_map <- vapply(dependencies, function(x) {
+    if (x$name %in% c('react', 'react-dom')) {
+      x$name <- 'dash_renderer'
+    }
+    dep_path <- paste(x$src$file, 
+                      x$script,
+                      sep = "/")
```
chriddyp: is this a file path? if so, should this be platform dependent or is there a platform independent way to construct platform paths? (on windows, it's `\`)

R/utils.R
```diff
@@ -327,4 +335,66 @@ valid_seq <- function(params) {
...
+  
+    # remove double slash and replace with single if present
```
chriddyp: when is there a double slash vs a single slash?

R/dash.R
```diff
@@ -257,17 +256,57 @@ Dash <- R6::R6Class(
...
+        
+        dep_pkg <- get_package_mapping(dist_path, 
+                                       keys$package_name,
+                                       c(private$dependencies_internal,
+                                                    private$dependencies,
+                                                    private$dependencies_user)
+                                       )
+        
+        dep_path <- system.file(dep_pkg$rpkg_path, 
+                                package = dep_pkg$rpkg_name)
```
chriddyp: from a security perspective, we have to be very careful that a malicious user won't be able to construct a URL in such a way that `dep_pkg` is a non-dash-component package (enabling that user to extract code from private packages _or_ arbitrary files on the file system). It looks like this is primarily handled by checking the URLs against `private$dependencies` in `get_package_mapping`.

R/dash.R
```diff
@@ -532,10 +545,7 @@ Dash <- R6::R6Class(
       ))
 
       # normalizes local paths and keeps newer versions of duplicates
```
chriddyp: do we expect that there will be duplicates? we don't do this resolve step on the python side - I suppose that a component author could make a mistake by including a JS file with the same name twice when specifying that component's dependencies.

R/dash.R
```diff
@@ -573,10 +583,11 @@ Dash <- R6::R6Class(
           </body>
         </html>',
         private$name,
-        render_dependencies(depsCSS, local = private$serve_locally),
+        render_dependencies(depsCSS, local = private$serve_locally, prefix=self$config$routes_pathname_prefix),
```
chriddyp: this should be `requests_pathname_prefix` - as the script tag is something that makes a request

R/dash.R
```diff
@@ -573,10 +583,11 @@ Dash <- R6::R6Class(
...
         to_JSON(self$config),
-        render_dependencies(depsScripts, local = private$serve_locally)
+        render_dependencies(depsScripts, local = private$serve_locally, prefix=self$config$routes_pathname_prefix)
```
chriddyp: similarly here

R/utils.R
```diff
@@ -129,31 +129,39 @@ render_dependencies <- function(dependencies, local = TRUE) {
...
+    }
+
+    # The following lines inject _dash-component-suites into the src tags,
+    # as this is the current Dash convention. The dependency paths cannot
+    # be set solely at component library generation time, since hosted
+    # applications should have the app name injected as well.
+    #
+    # This is essentially analogous to this codeblock on the Python side:
+    # https://github.com/plotly/dash/blob/1249ffbd051bfb5fdbe439612cbec7fa8fff5ab5/dash/dash.py#L207
+    if (is.null(prefix)) {
```
chriddyp: we should be able to remove this `is.null` check - by the time this function is called, we've already coerced `requests_pathname_prefix` from null to `/` in `resolve_prefix`

R/utils.R
```diff
@@ -129,31 +129,39 @@ render_dependencies <- function(dependencies, local = TRUE) {
...
+      dep[["script"]] <- paste0("/_dash-component-suites/",
+                                dep$name,
+                                "/",
+                                basename(dep[["script"]]))
+    } else {
+      dep[["script"]] <- paste0(sub("/", "", prefix),
```
chriddyp: I don't quite understand this `sub` - I believe we should be passing the prefix in directly here. if the prefix is `/my-app`, then that should be the start of this route, not `my-app`

R/utils.R
```diff
@@ -129,31 +129,39 @@ render_dependencies <- function(dependencies, local = TRUE) {
...
+                                "/",
+                                basename(dep[["script"]]))
+    } else {
+      dep[["script"]] <- paste0(sub("/", "", prefix), 
+                                "_dash-component-suites/", 
+                                dep$name,
+                                "/",
+                                basename(dep[["script"]]))
+    }
+    dep$src$file <- ""
```
chriddyp: what does this do / why is it necessary? could we add a comment?

R/utils.R
```diff
@@ -327,4 +335,66 @@ valid_seq <- function(params) {
...
\ No newline at end of file
+}
+
+resolve_prefix <- function(prefix, environment_var) {
+  if (!(is.null(prefix))) {
+    assertthat::assert_that(is.character(prefix))
+    
+    return(prefix)
+  } else {
+    prefix_env <- Sys.getenv(environment_var)
```
chriddyp: Is there way to check if an env variable is set in R? I don't expect it to come up in this case, but with this pattern it seems like it would be impossible for a user to set `DASH_REQUESTS_PATHNAME_PREFIX=""`, as we interpret `""` the same as the user didn't set it.
Shouldn't affect this case, but as we use this same type of config variable resolution for other variables, it might come up

R/utils.R
```diff
@@ -327,4 +335,66 @@ valid_seq <- function(params) {
...
+    if (prefix_env != "") {
+      return(prefix_env)
+    } else {
+      return("/")
+    }
+  }
+}
+
+get_package_mapping <- function(script_path, url_package, dependencies) {
+  package_map <- vapply(dependencies, function(x) {
```
chriddyp: I'm having a hard time following this function. Part of it is definitely my lack of R knowledge but it would be helpful if there were some comments that explained what these variables look like. For example:
```
# script_path is e.g. `/.../
# url_package is e.g. `...`
# dependencies is e.g. ...
# this function returns ...
```

R/utils.R
```diff
@@ -327,4 +335,66 @@ valid_seq <- function(params) {
...
+                      sep = "/")
+  
+    # remove double slash and replace with single if present
+    result <- c(pkg_name=x$package,
+                dep_name=x$name,
+                dep_path=gsub("//", replacement = "/", dep_path)
+    )
+  }, FUN.VALUE = character(3))
+  
+  package_map <- as.data.frame(t(package_map), stringsAsFactors = FALSE)
```
chriddyp: could we add a comment here? (why converting to a dataframe?)

R/utils.R
```diff
@@ -327,4 +335,66 @@ valid_seq <- function(params) {
...
+  
+  filename <- basename(script_path)
```
chriddyp: does `basename` handle cross-platform directories? (i.e. `\` on windows)

R/utils.R
```diff
@@ -327,4 +335,66 @@ valid_seq <- function(params) {
...
+  
+  pos_match <- grepl(paste0(filename, "$"), package_map[, "dep_path"]) &
```
chriddyp: a comment here explaining what these variables look like would help me understand this part of the code better

R/utils.R
```diff
@@ -129,31 +129,39 @@ render_dependencies <- function(dependencies, local = TRUE) {
...
+                                basename(dep[["script"]]))
+    } else {
+      dep[["script"]] <- paste0(sub("/", "", prefix), 
+                                "_dash-component-suites/", 
+                                dep$name,
+                                "/",
+                                basename(dep[["script"]]))
+    }
+    dep$src$file <- ""
     htmltools::renderDependencies(list(dep), src, encodeFunc = identity)
```
chriddyp: it would be good to get the query string parameter in here `?v=<version-of-the-package>` so that caching works. It sounded like this was tough to do with `htmltools`, so I'd recommend just constructing these yourself with `sprintf` or w/e
It's a simple string, it's either e.g. 
```
<script type="text/javascript" src="/_dash-component-suites/dash_core_components/dash_core_components.min.js?v=0.43.1"></script>
```
for javascript files
or e.g. 
```
<link rel="stylesheet" href="/_dash-component-suites/dash_core_components/tabs.css?v=0.43.1">
```
for css files

R/dash.R
rpkyle: OK, got it. Just curious, do these lines in `dash.py` remain for backward compatibility, then?

https://github.com/plotly/dash/blob/2a4c63ed6215973ca34262c61d6ce25947cf3643/dash/dash.py#L95

https://github.com/plotly/dash/blob/2a4c63ed6215973ca34262c61d6ce25947cf3643/dash/dash.py#L140

https://github.com/plotly/dash/blob/2a4c63ed6215973ca34262c61d6ce25947cf3643/dash/dash.py#L514

I was trying to reimplement this logic in R, down to the value of `2678400`.

R/dash.R
rpkyle: Ah, good point. There's nothing like this handler in Dash for Python, so it definitely seems :hocho:-able.

R/utils.R
rpkyle: Right, good question; I can't say with 100% certainty, but I think it might be OK here because of the way `system.file` is written:

https://github.com/wch/r-source/blob/5a156a0865362bb8381dcd69ac335f5174a4f60c/src/library/base/R/files.R#L201

...the line above calls `file.path`, which is platform-agnostic. From https://www.rdocumentation.org/packages/base/versions/3.5.1/topics/file.path:

>_Construct the path to a file from components in a platform-independent way._

I can test the code on my Windows VM and verify that this is the case.

R/utils.R
rpkyle: This problem will go away when I :hocho: the use of `htmltools` for generating the script tags. It seems to want to insert a leading slash in the tag, and I couldn't figure out how to remove it.

A far better solution is to do what you suggested earlier today, and manually generate the script tag as in Python. I'll take care of that and :hocho: these lines when they're no longer needed.

R/dash.R
rpkyle: Yeah, it's interesting that you mentioned this, because I realized we haven't discussed security considerations in DashR yet -- earlier today I was reviewing open issues, and saw this (completely unrelated, but on the topic of security):

https://github.com/plotly/dashR/issues/27

R/dash.R
rpkyle: This goes back to Nicolas's earlier example. I agree that it seems pretty far-fetched in practice, if we end up :hocho: references to `htmltools`, this will go away too.

R/utils.R
rpkyle: This definitely looks like an oversight on my part. Will fix, thanks for letting me know.

R/utils.R
rpkyle: Hmm. This is what happens when I'm coding too late at night. I had thought `renderDependencies` was inserting a leading slash, but that's obviously incorrect (and didn't make much sense). This will all go away if we hand-code the script tags.

Without the prefix, a simple test shows

```
> htmltools:::renderDependencies(dependencies = dashR:::.dashR_js_metadata(), srcType = "file")`

<script src="lib/react%4015.4.2/dist/react.min.js"></script>
<script src="lib/react%4016.2.0/umd/react.production.min.js"></script>
<script src="lib/react-dom%4015.4.2/dist/react-dom.min.js"></script>
<script src="lib/react-dom%4016.2.0/umd/react-dom.production.min.js"></script>
<script src="lib/dash-renderer%400.18.0/dash_renderer/dash_renderer.min.js"></script>
```

But, in the codeblock you referenced, if `prefix` is `'/my-prefix/'` and we change it instead to 

```
dep[["script"]] <- paste0("/my-prefix/", "_dash-component-suites/", 
                           dep$name, "/", basename(dep[["script"]]))
```

The resulting tag (double slash at head) is

`<script src="//my-prefix/_dash-component-suites/dash_renderer/react.min.js"></script>`

...whereas if `prefix` is `'/'` we see

`<script src="//_dash-component-suites/dash_renderer/react.min.js"></script>`

R/utils.R
rpkyle: Yes, I should definitely have added a comment here; this is a bit strange. Starting from

```
dep[["script"]] <- paste0(sub("/", "", prefix), "_dash-component-suites/", 
    dep$name, "/", basename(dep[["script"]]))
```

...then rendering the dependencies with `htmltools`:

`>  htmltools::renderDependencies(list(dep), src, encodeFunc = identity)`

...if we don't blank `dep$src$file`, we obtain

```
<script src="lib/react-dom@15.4.2/_dash-component-suites/dash_renderer/react-dom.min.js"></script>
```

versus

`<script src="/_dash-component-suites/dash_renderer/react-dom.min.js"></script>`

which is what I had thought we wanted.

R/utils.R
rpkyle: Definitely the way to go here, and would resolve some of the other issues you noted. Sorry for all the messy code, I'll fix this up and note the commit below.

R/utils.R
rpkyle: Yeah, the current approach is no good from that point of view. Yes, here's one possible approach:

```
> "DASH_REQUESTS_PATHNAME_PREFIX" %in% names(Sys.getenv())
[1] FALSE
> "HOME" %in% names(Sys.getenv())
[1] TRUE
> Sys.getenv("HOME")
[1] "/Users/rpkyle"
```

This may not work reliably across all platforms, we'd probably want to test:

>_On most platforms_ `Sys.getenv()` _will return a named vector giving the values of all the environment variables, sorted in the current locale. It may be confused by names containing = which some platforms allow but POSIX does not. (Windows is such a platform: there names including = are truncated just before the first =.)_

R/utils.R
rpkyle: OK, will add.

R/utils.R
rpkyle: Yes, if I understand correctly. From the `basename` man page:

> *Behaviour on Windows*
>
>On Windows this will accept either \ or / as the path separator, but `dirname` will return a path using / (except if on a network share, when the leading \\ will be preserved). Expect these only to be able to handle complete paths, and not for example just a network share or a drive.
>
> UTF-8-encoded path names not valid in the current locale can be used.

This post explains why I'm not able to get `basename` to work as I would like when I pass it `C:\testpath\bundle.js`, since I'm using R in a *nix environment:

https://stat.ethz.ch/pipermail/r-help/2012-August/333592.html

>Yes, basename(filepath) assumes that filepath represents path to a file on the OS that R is running on.  Hence, on Windows we get
```  
> basename("dir\\path")
  [1] "path"
  > basename("dir/path")
  [1] "path"
```
>because both `/` and `\` are valid separators (and invalid in file names)
>
>Linux allows `\` in a file name so we get
```
> basename("dir\\file")
[1] "dir\\file"
> basename("dir/file")
[1] "file"
```

R/utils.R
rpkyle: The statement generates a vector of `logical()` (Boolean) values which are used to filter the columns of the `package_map` object. 

In R, you can index a vector using a Boolean, which can be handy if the indices of elements to return are not always predictable:

```
> c(1,2,3,4)[c(FALSE, FALSE, TRUE, TRUE)]
[1] 3 4
```

The `paste0(filename, "$")` bit is used to create a regular expression requiring the filename to be the final characters in the string. Otherwise, you can run into trouble matching both `bundle.js` and `bundle.js.map` when you may only want `bundle.js`. I'll insert a succinct explanation within this codeblock.

R/utils.R
rpkyle: I suspect this is a holdover from an earlier version of the function in which strings were being coerced to factors and transposing the table was leading to some strange behaviour. I think I can safely replace this with

```suggestion
package_map <- t(package_map)
```

R/dash.R
chriddyp: Ah I see. Looks like this was added in https://github.com/plotly/dash/pull/387/. I'm not actually sure what the use case is. But sure, let's keep it in just for parity 👍

R/utils.R
chriddyp: Got it - let's use `file.path` here and anywhere that we're constructing file paths. That way, we're not dependent on other functions downstream handling our non-platform-dependent file strings correctly.

In python we do a similar thing with `os.path.join`

R/dash.R
chriddyp: it would be good if we could isolate some of this logic into a function that we could test

R/utils.R
chriddyp: Ah OK, that's odd. Seems like another case in favor of us just rendering these script tag strings ourselves with `sprintf`

R/dash.R
alexcjohnson: FWIW 2678400 is 31 days, so functionally equivalent to infinite most of the time.

R/dash.R
rpkyle: fixed in f0baaea64d062ee5b68013db83ef490ac3090b0b

R/dash.R
rpkyle: fixed in cc3ea048f7c2594e5efd137afcfb76324adba34d

R/dash.R
rpkyle: fixed in cc3ea048f7c2594e5efd137afcfb76324adba34d

R/dash.R
rpkyle: I tried to work with the information within the `keys` passed to the route, but it's not quite what we want. Using the URL syntax you provided above, I am able to :hocho: the messy `url` parsing block from L275:L278; thanks for noting this.

But `dep_pkg` doesn't line up with what is returned by `get_package_mapping`. Here are two dependencies as an example:

previous `dist_path`: `plotly-1.44.1.min.js`
new `dist_path`: `plotly-1.44.1.min.js`
previous `dep_pkg`: `deps/plotly-1.44.1.min.js`  
new `dep_pkg`: `dash_core_components_0` 

previous `dist_path`: `dash_core_components.min.js` 
new `dist_path`: `dash_core_components.min.js` 
previous `dep_pkg`: `deps/dash_core_components.min.js`
new `dep_pkg`: `dash_core_components`

rpkyle: fixed in 167079a9b747926794625f8eb51d6bb01756adf2

R/utils.R
rpkyle: fixed in bea8ffc1f058dab55533a55a888d04bfb55144d0

R/utils.R
rpkyle: fixed in 0569677c435c928878577db110b4036c9b0b12a8

R/utils.R
rpkyle: fixed in 0569677c435c928878577db110b4036c9b0b12a8

R/utils.R
rpkyle: fixed in 0569677c435c928878577db110b4036c9b0b12a8

R/utils.R
rpkyle: fixed in b34c4f841fc0fe45f9c527823a98de97c3778394

R/utils.R
rpkyle: fixed in 8362cb5

R/utils.R
rpkyle: fixed in 2368c34deac2785e609563890d8862cd25041410

rpkyle: I've opened a new issue to describe what's needed to fully :hocho: all references to `htmltools`: https://github.com/plotly/dashR/issues/54

R/dash.R
rpkyle: I've opened a issue related to this comment, given the potential consequences in terms of application security, and its critical nature:

https://github.com/plotly/dashR/issues/55

rpkyle:
Previously, we encountered an `error: subscript out of bounds` message in the R console when attempting to pass a `NULL` value to the callback handler. 

This is possible when using a component like `coreLocation`, in which an ID may be supplied without a default value. The original code:

```callback_args <- lapply(c(request$body$inputs, request$body$state), `[[`, 3)```

...has been replaced with:

https://github.com/plotly/dashR/blob/7fa2581d41e8a33c8966490f51ef0ccf371edfd8/R/dash.R#L244-L260

The new code handles inputs which may contain `NULL` values, which we wish to retain. In R, assigning `NULL` into a list element removes the element from a list, while appending `NULL` inserts into an existing list.

https://cran.r-project.org/doc/FAQ/R-FAQ.html#Others:

**Comment by rpkyle** [comment]
I've tried to address the most recent round of PR comments, and I think I've licked the `_dash-update-component` error Nadia mentioned. If it's possible to review and consider :dancer: this one, I'm ready to tackle other issues 🙂 

@chriddyp @alexcjohnson @NadiaTahiri @nicolaskruchten

File: R/internal.R [hunk 22]
```diff
@@ -1,63 +1,75 @@
...
+                                               head = NULL,
+                                               attachment = NULL,
+                                               package = "dashR",
+                                               all_files = FALSE),
+                                          class = "html_dependency"),
+                        `react-prod` = structure(list(name = "react",
+                                                      version = "16.2.0",
+                                                      src = list(href = "https://unpkg.com/react@16.2.0",
+                                                                 file = "lib/react@16.2.0"),
+                                                      meta = NULL,
```
chriddyp: It looks like this is just white space removal, but a few things strike me as odd about this code:
1. "react-prod" is react-16. This is different than the python side - we allow users to switch to react 16 optionally but it's not a "prod" vs "dev" distinction.
2. On the python side, we handle this in dash-renderer using the same structure as the rest of the component repos (https://github.com/plotly/dash-renderer/blob/1c7c75d3127a742093f59bd34ea23b9a84be1ac9/dash_renderer/__init__.py). I would be worried about DashR handling these versions independently of dash-renderer as we don't want DashPy to be distributed with a different version than DashR

This is non-blocking and not a high priority, but something to maybe break out into its own issue and consider down the line.

R/utils.R
```diff
@@ -129,31 +129,65 @@ render_dependencies <- function(dependencies, local = TRUE) {
...
+    # parameter for cache busting
+    
+    if (!is.null(dep$package)) {
+      # the gsub line is to remove stray duplicate slashes, to
+      # permit exact string matching on pathnames
+      dep_path <- gsub("//+",
+                       "/",
+                       paste(dep$src$file,
+                             dep$script,
+                             sep = "/")
```
chriddyp: If this is only used as a file path, could we use `file.path`? I recognize that it's being used below, but just to get into the habit of never constructing file paths by hand.

(unless this variable is also used as a script tag / url pathname? By reading the code below it didn't seem like it but I might've missed something)

R/utils.R
```diff
@@ -129,31 +129,65 @@ render_dependencies <- function(dependencies, local = TRUE) {
...
+      
+      full_path <- system.file(file.path(dep_path),
+                               package = dep$package)
+      
+      modified <- as.integer(file.mtime(full_path))
+    } else {
+      modified <- as.integer(Sys.time())
+    }
+    
+    # we don't want to serve the JavaScript source maps here
```
chriddyp: maybe add a comment like "until we have full support for a debug=TRUE mode"?

R/utils.R
chriddyp: can we get into the habit of using `file.path` here?

chriddyp: Or, since you are replacing `/` below, maybe you don't want to use `file.path` here because that could return `/` on windows? If that's the case, could add a comment to elaborate?

R/internal.R
rpkyle: Thanks for this comment -- this is among the last of the code/objects still retained from the "old" version of `dashR`. This function is written by the "transpiler" code when component libraries are generated, but I modify it directly as needed.

I'll open an issue to resolve this discrepancy and then link here.

R/utils.R
```diff
@@ -327,4 +361,79 @@ valid_seq <- function(params) {
...
+  #       to avoid duplicating dependencies in the package_map
+  package_map <- vapply(unique(dependencies), function(x) {
+    if (x$name %in% c('react', 'react-dom')) {
+      x$name <- 'dash-renderer'
+    }
+    dep_path <- paste(x$src$file, 
+                      x$script,
+                      sep = "/")
+  
+    # remove n>1 slashes and replace with / if present
```
chriddyp: could you prefix this comment with something like "htmltools sometimes includes multiple `//` - replace these with `/`". 
That way, the reader knows that `//` is coming from htmltools and not necessarily the dash-renderer API. In the future if you refactor out `htmltools`, then the reader will know that this `//` won't need to be handled because `htmltools` was the source of the issue.

R/utils.R
rpkyle: Good call. I could open an issue for this as well, I'll have a look at what's needed to approach parity based on what's currently available in Dash for Python.

R/utils.R
rpkyle: I've started the process of :hocho: `htmltools`, managed to remove nearly all references to this package, but not quite there yet. I opened an issue earlier this week which attempts to enumerate all the spots in the code and comments that are involved:

https://github.com/plotly/dashR/issues/54

chriddyp: Yeah - My suggestion here is mostly just to get in the habit of writing these types of commits - if there is something in the code that looks odd, it's good to get in the habit of explaining _why_ we have to do stuff like this and where the problem originated.

In this case, everyone involved in this PR is now aware of the issue and you will be the one refactoring it shortly. However, imagine the case where this refactor might happen 6 months from now _by someone else_. If they didn't know that this `//` was originating from `htmltools`, then they might not feel comfortable removing this block of code. Perhaps they might think that `dash-renderer` (a 3rd party piece of code) _sometimes_ introduces a `//` in certain scenarios, scenarios that might not necessarily be tested in integration tests.

So, it's not strictly necessary in this scenario where we plan on fixing this up in a couple of weeks, but it's a good habit to get into and a healthy way to start thinking about comments

R/utils.R
rpkyle: fixed in 3a61d023c1603b9d6d1f6303351b620cbb5262ac

R/utils.R
rpkyle: fixed in 3a61d023c1603b9d6d1f6303351b620cbb5262ac

R/utils.R
rpkyle: fixed in b6af99b6e58ce65791ff471c63b43125632d9cdd

R/utils.R
rpkyle: For sure, decisions like these seem apparent now that I'm staring at the same code over and over again, but for anyone else (or me even a year from now), statements like this are hard to comprehend in total isolation.

I'll be a bit more diligent about making these sorts of commits, it benefits everyone from the author, to the reviewer, and even savvy users who decide to take a look under the hood.

rpkyle: fixed in b77c94d57625d31e9914511943ea730e6fc4e86d

Issue #56, opened 2019-03-11
Address bug in remote CSS dependency handling


rpkyle:
https://github.com/plotly/dashR/pull/53 introduced a `get_package_mapping` function to construct a map matching `keys$package` from the `_dash-component-suites` handler to a (currently loaded) R package which supplies the dependency (if present).

While JavaScript dependency handling appears to work as intended, the update was not tested against remote CSS dependencies. This should be resolved before asset serving is implemented.

The crux of the issue is that `get_package_mapping` currently requires that four pieces of information are available:
- a dependency name
- the dependency "package" name from the URL (which is generally ≠ to the R package name)
- a list of dependencies in `htmltools` format
- the R package name, registered within each element of the dependency list described above

A PR (https://github.com/plotly/dashR/pull/57) has been proposed to handle this potential issue, by asserting that dependencies have valid name, URL package, and R package information before constructing the package map.

PR #57, opened 2019-03-11
Validate list of dependencies prior to constructing package_map

rpkyle:
This PR proposes introducing a function called `clean_dependencies` to help ensure that the list of dependencies passed to `get_package_mapping` contains no elements which are missing one or more of the following:
- a dependency name
- the dependency "package" name from the URL (which is generally ≠ to the R package name)
- a list of dependencies in `htmltools` format
- the R package name, registered within each element of the dependency list described above

Closes https://github.com/plotly/dashR/issues/56.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -176,7 +176,7 @@ render_dependencies <- function(dependencies, local = TRUE, prefix=NULL) {
     # until we are able to provide full support for debug mode,
     # as in Dash for Python
     if ("script" %in% names(dep) && tools::file_ext(dep[["script"]]) != "map") {
-      if !(is.null(dep$src$href)) {
+      if (!(is.null(dep$src$href))) {
         html <- sprintf("<script src=\"%s\"></script>", dep$src$href)
       } else {
         dep[["script"]] <- paste0(prefix,
--- a/R/utils.R
+++ b/R/utils.R
@@ -163,7 +163,6 @@ render_dependencies <- function(dependencies, local = TRUE, prefix=NULL) {
       dep_path <- gsub("//+",
                        "/",
                        dep_path)
-      )
       
       full_path <- system.file(dep_path,
                                package = dep$package)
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -292,6 +292,33 @@ filter_null <- function(x) {
...
+# this attempts to avoid cryptic errors produced by
+# get_package_mapping, which requires three parameters:
+# -- the script name (i.e. x$script below)
+# -- the package name from the URL (i.e. x$package)
+# -- the list of dependencies (i.e. deps)
+#
+# within get_package_mapping, x$package is also required,
+# so deps missing it here are assigned NULL and then
+# filtered out by the subsequent vapply statement
+assert_valid_dependencies <- function(deps) {
```
alexcjohnson: `assert` says to me that this function performs a test - which would then either throw an error or return a boolean. So I'd call this something like `clean_dependencies` instead.

rpkyle: I agree, but I realized that I should add a check for non-stylesheet deps that ensures the essential details are present, and throws an error if they aren't.

Since I will be adding at least one test, should I still rename this function?

alexcjohnson: It's more about what the function returns than anything else - you can have tests throwing an error inside any function, but if it's *name* is `assert_*` then it should have either no return value or it should return a boolean:
`all_deps_ok = assert_valid_dependencies(deps)`
Above what you effectively have is:
`valid_dep_list = assert_valid_dependencies(deps)`
where the verb `assert` just doesn't match the return value.

rpkyle: fixed in f85a8da0abfacabc85c2cfd865dd8d275de8d073

rpkyle:
@nicolaskruchten suggested not :hocho: the remote URLs for the JavaScript dependencies, so I've removed that (last) commit from the stack.

Issue #58, opened 2019-03-12
Provide support for external_stylesheets when initializing a DashR app


rpkyle:
Dash for Python provides a simple, convenient method for app developers to insert URLs for external stylesheets into the page header. DashR currently lacks this functionality; implementing this requires accepting an `external_stylesheets` argument, which will be a list of URLs, as in Dash for Python:

https://github.com/plotly/dash/blob/e255ba0ff3ce5e3080b596ee04f9e5c902a07953/dash/dash.py#L97

Code should be written which iterates over this list and uses `sprintf` to prepare stylesheet tags, which we can then insert into the page header, as in Python:

https://github.com/plotly/dash/blob/e255ba0ff3ce5e3080b596ee04f9e5c902a07953/dash/dash.py#L418-L427

PR #59, opened 2019-03-12
Support external_stylesheets and handle remote CSS as in Dash for Python

rpkyle:
This PR aims to eliminate the current method of providing remote CSS in DashR, and adopts the same approach currently used in Dash for Python:

https://github.com/plotly/dash/blob/e255ba0ff3ce5e3080b596ee04f9e5c902a07953/dash/dash.py#L97
https://github.com/plotly/dash/blob/e255ba0ff3ce5e3080b596ee04f9e5c902a07953/dash/dash.py#L418-L427

A function named `generate_css_dist_html` is added to `utils.R`, and called from `private$index`. The function takes a URL string as input, and returns a script tag ready to insert into the page header. In `dash.R`, an `external_stylesheets` parameter defaults to `NULL` but accepts a list of URLs.

Closes https://github.com/plotly/dashR/issues/58.

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -27,10 +27,15 @@
 #'   `routes_pathname_prefix` \tab \tab a prefix applied to the backend routes.\cr
 #'   `requests_pathname_prefix` \tab \tab a prefix applied to request endpoints
 #'   made by Dash's front-end.\cr
+#'   `external_stylesheets` \tab \tab An optional list of valid URLs from which
+#'   to serve CSS for rendered pages.\cr
 #'   `suppress_callback_exceptions` \tab \tab Whether to relay warnings about
 #'   possible layout mis-specifications when registering a callback. For an
 #'   example of when it is ok to suppress these callbacks, see
-#'   `runTutorial("urls-part-2.R")`
+#'   `runTutorial("urls-part-2.R")`\cr
+#'   `components_cache_max_age` \tab \tab An integer value specifying the time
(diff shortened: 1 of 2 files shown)


R/dash.R
```diff
@@ -27,10 +27,15 @@
...
 #'   `requests_pathname_prefix` \tab \tab a prefix applied to request endpoints
 #'   made by Dash's front-end.\cr
+#'   `external_stylesheets` \tab \tab An optional list of valid URLs from which
+#'   to serve CSS for rendered pages.\cr
 #'   `suppress_callback_exceptions` \tab \tab Whether to relay warnings about
 #'   possible layout mis-specifications when registering a callback. For an
 #'   example of when it is ok to suppress these callbacks, see
-#'   `runTutorial("urls-part-2.R")`
+#'   `runTutorial("urls-part-2.R")`\cr
+#'   `components_cache_max_age` \tab \tab An integer value specifying the time
```
nicolaskruchten: where did this extra bit of docs come from?

nicolaskruchten:
💃 
You could probably just squash this PR :)

R/dash.R
rpkyle: I failed to update the help to describe `components_cache_max_age` before merging an earlier PR:

https://github.com/plotly/dashR/pull/53

I caught this when I was documenting the `external_stylesheets` argument, and attempted to sneak it by you...and failed.

PR #64, opened 2019-03-21
[WIP] Initial implementation of asset serving in DashR

### LINKED ISSUE #46: Asset serving
There's a surprisingly deep and complicated pipeline behind the generation of the component `<script>` tags that Dash drops into the `index.html` which it returns to clients on first load.

Minimally, DashR needs to serve `script` tags for *locally served* JS bundles that match the component libraries that are loaded in the initial layout.

This is on the critical path for #45

rpkyle:
This PR proposes to provide support for locally serving assets from a specified assets folder. As in Dash for Python, the following features would be implemented upon merging these commits (`/assets` below refers to whichever folder is specified in `assets_folder`):

- Assets folder auto resolution (`/assets`) at the root of the project by default.
- Assets folder may be passed as parameter, customizing the path as desired
- JS/CSS files in `/assets` are served in the index
- An `assets_ignore` filter permits excluding assets matching specified regex
- Assets are cached by the browser, and uncached when modified
- All files in `/assets` are served statically
- The generated assets URL respects routes/requests pathname prefix
- Easy retrieval of assets URLs is possible: `get_asset_url(asset)` returns a ready-to-insert URL string 
- A `favicon.ico` is served as the title icon of the page, if in the root directory of `/assets`

There ~are~ is at least ~two~ one item~s~ that ~have~ has not yet been implemented, but which ~are~ is desirable:

- ~A `favicon.ico` is not currently served as the title icon of the page~
- Automatic compression of served assets is not currently supported, but should be feasible

While the above features have been tested, formal tests asserting that these elements are provided should be be written. An example of the Python code which performs this task is available here:

https://github.com/plotly/dash/blob/master/tests/test_resources.py

Closes #46.


R/dash.R
```diff
@@ -6,7 +6,10 @@
 #' app <- Dash$new(
 #'   name = "dash",
 #'   server = fiery::Fire$new(),
-#'   static_folder = NULL,
+#'   static_folder = 'static',
```
alexcjohnson: Can we get rid of `static_folder` now, so we don't create a breaking change down the line like we're going to have for dashpy? https://github.com/plotly/dash/issues/469#issuecomment-450355079

R/dash.R
```diff
@@ -292,28 +319,79 @@ Dash <- R6::R6Class(
...
-                            sprintf('public, max-age=%s', 
+        response$set_header('Cache-Control',
+                            sprintf('public, max-age=%s',
                                     components_cache_max_age)
                             )
         response$type <- get_mimetype(filename)
         TRUE
       })
 
+      dash_assets <- paste0(self$config$routes_pathname_prefix, private$assets_folder, "/*")
```
alexcjohnson: shouldn't this one be `assets_url_path` instead of `assets_folder`? Perhaps with its leading `/` stripped...

https://github.com/plotly/dash/blob/69b112df8153ad416fb279944e686cec7dc524a1/dash/dash.py#L158-L161

R/dash.R
rpkyle: Yes, let’s :hocho: this.

R/dash.R
rpkyle: I suspect that the end result might be the same, but I agree with you, logically I should use the URL here. Will fix.

alexcjohnson: They need not be the same though... we might have:
`routes_pathname_prefix='my/app'`
`assets_url_path='/assets'`
`assets_folder='some/local/path'`
(Incidentally I don't know why we include a leading slash in `assets_url_path` if we're just going to strip it... dash does it this way too so not really a question for you, I just don't understand it)

Then suppose we have a file `some/local/path/subfolder/pink.css`, we want that to be served at a URL like `http://localhost:8080/my/appassets/subfolder/pink.css?123456`

I'd think then that the easiest way to match these up would be (while walking the assets folder) to make a mapping of the url full path (stripping off the host and query string) to the file local path, in this case `'my/appassets/subfolder/pink.css' -> 'some/local/path/subfolder/pink.css'`

rpkyle: Ah, you're right - good point. I'll ensure that the URL path and assets folder can be as distinct as needed. I've already written regex to parse and excise the host information and `?...` bits, so that part should be :cake:.

R/dash.R
rpkyle: fixed in 16c4efa9d9c08a0b07c3b6483bb0ba220f589303

R/dash.R
rpkyle: fixed in fd821fd4bcef02e2bba25afe8708cef55cbdc2d1

rpkyle:
@alexcjohnson Let me know if this is looking a bit better; I've also added support for a `favicon.ico` in 74392b1.

R/dash.R
```diff
@@ -149,26 +157,21 @@ Dash <- R6::R6Class(
...
-        # these should all be directories, right?
-        if (!all(dir_exists(local_path))) {
-          warning(
-            "One or more of the following paths is not a directory, '%s'",
-            paste(static_folder, collapse = "', '"),
+      if (!is.null(private$assets_folder)) {
+        if (!(dir.exists(private$assets_folder))) {
+          warning(sprintf(
+            "The supplied assets folder, '%s' could not be found in the project directory.",
+            private$assets_folder),
```
alexcjohnson: dashPy doesn't look like it has this check. I like the check, but it'll cause problems for people who start an app with the default `assets` here but they don't have (or need) any assets.

Simplest solution I see is to ignore this warning if the folder is the default. It would be even better if we could ignore the error if the user didn't provide a folder name - meaning still give the warning if they provided a value that happens to match the default - but I don't think it would be a good tradeoff to drop the default from the function signature and lose that automatic documentation.

rpkyle: Ah, good point. (I should just have a template for your PR reviews where this is the default response.)

So, ultimately what we want is to keep this warning, but
- _don't display it_ when `assets_folder` is set to `assets`, the default, and the folder cannot be found
- _display the warning_ when `assets_folder` is user-supplied, ≠ `assets`, and cannot be found

Did I get that right?

PR #74, opened 2019-03-31
Remove dependencies_* methods in Dash class

### LINKED ISSUE #65: Make dependencies_get and dependencies_set private
DashR will soon support inlining external stylesheets, external JavaScripts, as well as asset serving from a local folder. The `dependencies_get` and `dependencies_set` methods should be made private, and potentially removed in the future, as this interface lacks parity with Dash for Python.

rpkyle:
This PR proposes to ~relocate~ remove three methods:
- `dependencies_set`
- `dependencies_get`
- `dependencies_get_internal`

~to the `private` field of the Dash R6 class.~ Closes #65.


rpkyle:
@TahiriNadia If you could take a look, would be grateful. Thanks in advance! 🙂

rpkyle:
@TahiriNadia Please review this PR, and confirm that removing these methods does not cause any issues with loading DashR apps. It should not cause any issues, but we'll need to be certain. Thanks 🙂

TahiriNadia:
The update gives me an error 
```
Error in eval(parse(text = example.ready.for.eval)) :
  attempt to apply non-function
Calls: source ... withVisible -> eval -> eval -> <Anonymous> -> eval -> eval
Execution halted
```

TahiriNadia:
```eval(parse(text=example_string))``` is used in `utils.R` file
```
LoadAndDisplayComponent <- function(example_string) {
  return(htmlDiv(list(
    htmlDiv(
      children=dccSyntaxHighlighter(example_string),
      className='code-container'
    ),
    htmlDiv(
      className='example-container',
      style = list("overflow-x" = "initial"),
      children=eval(parse(text=example_string))
    )
  )))
}
```

rpkyle:
> `eval(parse(text=example_string))` is used in `utils.R` file

@TahiriNadia The `utils.R` script is not part of the DashR package; there are no `eval(parse` calls in DashR itself.

Could you try again, but instead of the DashR docs, try a simpler test application or two? Remember that this branch is old enough that it still uses `layout_set` instead of `layout`.

I just re-ran my code, and there are no issues on my side when removing these methods (they aren't essential for writing a basic app, so it makes sense). Thanks in advance for your help.

rpkyle:
I think I see what's happening here. The methods have been :hocho:, but the docs need to be regenerated to reflect this. 

I believe it makes sense to :hocho: these, they aren't necessary anymore -- we should use `assets`, `external_stylesheets`, and `external_scripts` to serve dependencies, as this is idiomatic for Dash in Python as well.

Since I'm not using this API either, it feels a bit pointless to privatize it at this point. This is the old way, we're pretty close to abandoning `htmltools` altogether, and it seems likely they'll eventually fail. My plan post-release is to refactor dependency management to more closely mimic the Python side of things anyhow.

@alexcjohnson @nicolaskruchten May I excise these methods :point_up: ?

alexcjohnson:
Yes, 🔪

TahiriNadia:
I agree 🔪 if not needed. You can 💃

Issue #86, opened 2019-04-28
Support for debugging DashR errors needs improvement


rpkyle:
Currently, DashR lacks adequately informative error messages; errors such as the one seen here are likely to be very hard for developers (and app users) to interpret:

![image-2](https://user-images.githubusercontent.com/9809798/56858610-f6c34700-694a-11e9-89d2-5012cbaad6cd.png)

Ideally, we would offer several enhancements to improve the debugging experience when errors are thrown while in `debug` mode:
- return a full call stack, or a partial one which has been "pruned" of internal function calls either related to the error itself or initial configuration tasks prior to invoking DashR-related functions
- catch the error and instead display it as a warning, which may enable more productive debugging once support for interactive breakpoints (accessed via the callback graph) is implemented; currently an error "freezes" the DashR app

It would also be nice to
- return line numbers and function names for errors; this is complicated by R's current handling of `srcref` attributes (see below).

https://stat.ethz.ch/R-manual/R-devel/library/base/html/srcfile.html
http://r.789695.n4.nabble.com/Is-there-a-way-to-get-R-script-line-number-td4731127.html
https://journal.r-project.org/archive/2010-2/RJournal_2010-2_Murdoch.pdf

rpkyle:
The majority of these improvements were made in #87. Line numbers will addressed in an upcoming PR, but for the time being the basic support for improved debugging is now in place.

PR #87, opened 2019-04-28
[WIP] Add support for full and partial call stacks, improved error handling in DashR

### LINKED ISSUE #16: Tools for debugging components
It's hard to know when/why a component is 'mis-specified'. Furthermore, my current debugging process is not ideal (`app$layout_set(component); app$layout_get(); app$run_server()`). Some ideas for improvements:

1. Printing a component could spin up a simple dash app with just that component.
2. Provide a 'debugging mode/option' where client side errors are listed in a pop-up (the current generic `Error loading dependencies` message is not helpful and most folks won't know to open their browser's dev tools)
3. Automatically type check arguments in R using the components `propTypes` definition

rpkyle:
This PR proposes multiple improvements to error handling and debugging in DashR:
- Adds a `debug` mode to DashR apps
- Introduces full/partial call stacks if app is served via `app$run_server(debug=TRUE)`
- While in debug mode, catches errors as warnings, allowing app execution to continue
- Introduces logic to return an HTML version of the call stack and error message for use by `dash-renderer` and the recently released in-browser dev tools

Here is an example of an error before the proposed changes:

Screenshot 2019-04-28 00 46 52

...and after:

Screenshot 2019-04-28 00 47 07

Eventually, it should also be possible to conditionally implement line number support for R code loaded via `source()`.

Closes #16 and #86.


alexcjohnson:
Nice, that's certainly an improvement!

> * Introduces logic to return an HTML version of the call stack and error message for use by `dash-renderer` and the recently released in-browser dev tools

Does that work now? What does it look like? Also, am I reading it right that it's returning code 200 in the new version? Can we change it back to 500 (which is also what dash-py uses for errors)?

R/dash.R
```diff
@@ -467,12 +469,16 @@ Dash <- R6::R6Class(
                           port = Sys.getenv('DASH_PORT', 8050), 
                           block = TRUE, 
                           showcase = FALSE, 
+                          pruned = TRUE,
```
alexcjohnson: Maybe a more verbose name for this flag? `pruned_errors` perhaps?

rpkyle: Good idea, I'll make this change.

rpkyle: fixed in 35dedac

rpkyle:
> Nice, that's certainly an improvement!

😌
 
> > * Introduces logic to return an HTML version of the call stack and error message for use by `dash-renderer` and the recently released in-browser dev tools
> 
> Does that work now? What does it look like? Also, am I reading it right that it's returning code 200 in the new version? Can we change it back to 500 (which is also what dash-py uses for errors)?

Right, 500 would make more sense. For HTML version of the call stack, I'll need to emulate what `werkzeug` is doing on the Python side, no? Since `fiery` lacks that sort of R debugging integration, I believe this would be best accomplished by sending a response up to the renderer with the proper headers, but I'm not sure.

Any hints you can offer there would be much appreciated 🙂. I _think_ the frontend code that handles this sort of thing is here:

https://github.com/plotly/dash-renderer/blob/master/src/components/error/FrontEnd/FrontEndError.react.js

rpkyle:
Also, it looks like Carson raised a related issue with Thomas Lin Pedersen, but I don't think it went anywhere:

https://github.com/thomasp85/fiery/issues/34

alexcjohnson:
> Right, 500 would make more sense. For HTML version of the call stack, I'll need to emulate what `werkzeug` is doing on the Python side, no? Since `fiery` lacks that sort of R debugging integration, I believe this would be best accomplished by sending a response up to the renderer with the proper headers, but I'm not sure.

I think all you need is to send back the error as `response$body`, and set `response$status = 500L`. The error should be an HTML string, but that could initially just be a super simple wrapping like `'<!DOCTYPE HTML><html><body><pre>' + error_with_stack_trace + '</pre></body></html>'` and we can play with styling it nicely in another PR.

R/dash.R
```diff
@@ -255,6 +254,9 @@ Dash <- R6::R6Class(
           }
         }
 
+        # set the callback context associated with this invocation of the callback
+        private$callback_context_ <- setCallbackContext(request$body)
```
alexcjohnson: This looks great. I think all you need to do to finish off the callback context part of this PR is set it back to `NULL` after the `do.call` below, and have `callback_context` throw an error if it finds `NULL`.

rpkyle: fixed in 6fc1cac2e3506ba67ead79b496b51127ba7314e0

alexcjohnson: Oh sorry - what I meant was, down here when you access `callback_context_`

https://github.com/plotly/dashR/pull/87/files#diff-1881af3720f8f49fd3a44e51c9fb5a0dR489

At that point, if it's `NULL`, that means you're not in a callback so you shouldn't be able to access it, so throw an error.

rpkyle: fixed in 96e4672

rpkyle:
> > Right, 500 would make more sense. For HTML version of the call stack, I'll need to emulate what `werkzeug` is doing on the Python side, no? Since `fiery` lacks that sort of R debugging integration, I believe this would be best accomplished by sending a response up to the renderer with the proper headers, but I'm not sure.
> 
> I think all you need is to send back the error as `response$body`, and set `response$status = 500L`. The error should be an HTML string, but that could initially just be a super simple wrapping like `'<!DOCTYPE HTML><html><body><pre>' + error_with_stack_trace + '</pre></body></html>'` and we can play with styling it nicely in another PR.

fixed in cee5376

Issue #91, opened 2019-06-10
error for adding external Javascript


z267xu:
1. If we set external Javascript in 
`Dash$new(external_scripts = "your_own_js")`,
we would have error like:

_"Error in as.character(function (description, open = "", blocking = TRUE,  : cannot coerce type 'closure' to vector of type 'character'"_

The error comes from `generate_js_dist_html` function in `dashR` package utils file. The reason is the typo in `sprintf`, which should be **href** instead of **url**.

https://github.com/plotly/dashR/blob/6bc44b5c4203b03bd5ef04ab5dd24b23bf9b987b/R/utils.R#L513-L520

2. Also if we mix resources with attributes like `src`, `integrity`, `crossorigin`, etc, like:

`external_scripts <- list(list(`
`src = 'https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.10/lodash.core.js', `
`integrity = 'sha256-Qqd/EfdABZUcAxjOkMi8eGEivtdTkh3b65xCZL4qAQA=', `
`crossorigin = 'anonymous'))`

`Dash$new(external_scripts = external_scripts)`

we may have error like

"warning: the condition has length > 1 and only the first element will be used from if (grepl("^(?:http(s)?:\\/\\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$",
error: values must be length 1,
 but FUN(X[[2]]) result is length 4 from vapply(self$config$external_stylesheets, generate_css_dist_html,"

I guess we should manipulate input `href` before we call function `generate_js_dist_html()` and `generate_css_dist_html()`

https://github.com/plotly/dashR/blob/6bc44b5c4203b03bd5ef04ab5dd24b23bf9b987b/R/utils.R#L492-L532

@rpkyle

rpkyle:
@z267xu Thanks for this report, very helpful.

(1) I had run into the `href` vs. `url` issue previously, and I think I resolved it locally but the commits hadn't been pushed yet. I'll open a PR and aim to merge the changes into the `dev` branch.

(2) These attributes were added to Dash for Python in August 2018, so I should've incorporated them in Dash for R as well! Thanks for noting this. I think it might merit an issue so I don't lose track of it, will try to fix in the next week. It'll get added to the same PR I use for (1).

[...]

PR #94, opened 2019-06-13
Import CSS deps into package map

rpkyle:
This PR proposes to implement registration of CSS assets within R packages via the package map. 

Previously, importing via `external_stylesheets` and `assets` was supported, and in-package CSS loading was incompletely supported.

Partly addresses https://github.com/plotly/dashR/issues/91.


R/utils.R
```diff
@@ -157,8 +157,9 @@ render_dependencies <- function(dependencies, local = TRUE, prefix=NULL) {
     if (!is.null(dep$package)) {
       # the gsub line is to remove stray duplicate slashes, to
       # permit exact string matching on pathnames
-      dep_path <- file.path(dep$src$file,
-                            dep$script)
+      dep_path <- ifelse(!is.null(dep$script),
+                         file.path(dep$src$file, dep$script),
+                         file.path(dep$src$file, dep$stylesheet))
```
alexcjohnson: 🌴 
https://github.com/plotly/dashR/pull/94/files#diff-cbb7235afee10c3646161358a7655b0aR465

R/utils.R
```diff
@@ -192,10 +193,23 @@ render_dependencies <- function(dependencies, local = TRUE, prefix=NULL) {
                                                                        dep[["stylesheet"]], 
                                                                        sep="/"))
     } else if ("stylesheet" %in% names(dep) & src == "file") {
+      dep[["stylesheet"]] <- paste0(path_prefix,
+                                "_dash-component-suites/",
+                                dep$name,
+                                "/",
+                                basename(dep[["stylesheet"]]),
+                                sprintf("?v=%s&m=%s", dep$version, modified))
```
alexcjohnson: 🐄 seems funny to combine `sprintf` and `paste0` - I guess that's because `modified` isn't a string, and I see this is used above too... but it would be easier to read if we pick one for any given concatenation.

R/utils.R
```diff
@@ -192,10 +193,23 @@ render_dependencies <- function(dependencies, local = TRUE, prefix=NULL) {
...
+      
       if (!(is.null(dep$version))) {
-        html <- sprintf("<link href=\"%s?v=%s\" rel=\"stylesheet\" />", file.path(dep[["src"]][["file"]],
-                                                                                  dep[["stylesheet"]]),
-                        dep$version)        
+        if(!is.null(dep$package)) {
+          html <- sprintf("<link href=\"%s?v=%s\" rel=\"stylesheet\" />", file.path(dep[["stylesheet"]]),
+                          dep$version)
```
alexcjohnson: why is `dep$version` appearing both here and above where you define `dep[["stylesheet"]]`?

R/utils.R
rpkyle: 🙈fixed in d9d66c13062da253b0f77bb131126f64f12e8516

R/utils.R
rpkyle: :facepalm: fixed in d9d66c13062da253b0f77bb131126f64f12e8516

R/utils.R
rpkyle: 🐪 fixed in d9d66c13062da253b0f77bb131126f64f12e8516

R/utils.R
```diff
@@ -545,9 +548,12 @@ generate_css_dist_html <- function(href,
 generate_js_dist_html <- function(href, 
                                   local = FALSE,
                                   local_path = NULL,
-                                  prefix = NULL) {
+                                  prefix = NULL,
+                                  as_is = FALSE) {
   if (!(local)) {
-    if (grepl("^(?:http(s)?:\\/\\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$", href, perl=TRUE)) {
+  if (grepl("^(?:http(s)?:\\/\\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$",
```
rpkyle: This is just a formatting change to decrease the width of the line.

R/dash.R
```diff
@@ -481,6 +479,13 @@ Dash <- R6::R6Class(
           state=state,
           func=func
         )
+      
+      # # register the callback_context elements
+      # private$callback_context_[[paste(output$id, output$property, sep='.')]] <- list(
+      #   states=setCallbackContext(state),
+      #   triggered=list(),
+      #   inputs=getContextElements(inputs)
+      # )
```
alexcjohnson: what's this commented block?

rpkyle: When I initially implemented `callback_context`, I thought it would make sense for the object holding the details to exist before the callback was fired. I've since come to the conclusion that `callback_context` should be `NULL` until the callback is invoked. 

I had not yet removed the lines because it could simplify formatting the context if the object existed already, but I identified a workaround and users report that `callback_context` is fully functional. 

I've removed the inactive statements.

fixed in ba8ecf3

rpkyle:
> 💃 One more down ✅

:pray:

PR #97, opened 2019-06-19
Crawl package namespaces for CSS/JS deps metadata

### LINKED ISSUE #50: Address potential JS dependency loading issue for multipage DashR apps
@chriddyp raised an interesting weakness in the current dependency loading strategy for DashR (via calling the Python-generated JS metadata function (e.g. `.dashHtmlComponents_js_metadata()`).

In a multipage DashR app, it's possible that some dependencies may load properly for the first page, while any new components whose JS dependencies were not initially sourced may fail to load correctly.

There is a clever workaround in Dash for Python, but one does not currently exist in DashR.

I propose that we attempt the following strategy:
- 1. Enumerate the installed packages which have an R package dependency on `dashR`
- 2. Determine which of the dependent packages are currently attached (loaded)
- 3. Check to see if the packages in (2) contain an internal JS metadata function, if so, tack its dependencies on to those already registered by DashR.

Here's an example to demonstrate what I'm proposing (`.packages()` invisibly returns the list, so it's printed here):
```> dependsOnPkgs("ggplot2")
[1] "crosstalk"  "plotly"   
> print(.packages())
[1] "plotly"             "ggplot2"            "tools"              "dashRtutorial"      "magrittr"           "dashCoreComponents" "dashHtmlComponents"
[8] "dashR"              "stats"              "graphics"           "grDevices"          "utils"              "datasets"           "methods"           
[15] "base"  
> deps <- dependsOnPkgs("ggplot2")
> attached <- .packages()
> attached[attached %in% deps]
[1] "plotly"
```

I'm bumping this towards the top of the todo list because it could interfere with DashR docs.

@NadiaTahiri @T4rk1n

rpkyle:
As originally noted in #50, the current strategy of crawling the Dash layout to identify packages that supply CSS and JavaScript dependencies is potentially problematic. 

If the initial layout does not contain any components from a given library, their associated assets will not be loaded. This PR proposes to modify the dependency resolution process somewhat:
- retrieve the namespace information for each currently attached (loaded) package
- subset the list of objects using regex such that only functions whose name matches `.dash*_js_metadata` are retained (this omits the Dash for R internal metadata, to prevent loading these assets twice)
- as before, retrieve the contents of the metadata function and append to the dependency list for loading

Thanks to @sacul-git for reporting the initial manifestation of this issue, and for supplying a reproducible example for testing purposes.

Closes #50.

@alexcjohnson @chriddyp

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -594,9 +594,11 @@ Dash <- R6::R6Class(
         )
       }
 
-      # load package-level HTML dependencies
-      pkgs <- unique(layout_flat[grepl("package$", layout_nms)])
-      deps_layout <- lapply(pkgs, function(pkg) {
+      # load package-level HTML dependencies from attached pkgs
+      metadataFns <- lapply(.packages(), getDashMetadata)
+      metadataFns <- metadataFns[lengths(metadataFns) != 0]
+      
+      deps_layout <- lapply(metadataFns, function(dep) {
         # the objective is to identify JS dependencies
         # without requiring that a proprietary R format
         # file is loaded at object initialization to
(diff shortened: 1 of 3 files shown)


rpkyle:
Initial feedback from users has been positive; the fix appears to work as intended.

PR #96, opened 2019-06-17
Modifications required for R package release

rpkyle:
This PR proposes several modifications in support of the initial major release of Dash for R:
- `assert_valid_wildcards` now inspects `names(args)` rather than `args`
- `filter_null` has been renamed to `dash_filter_null`
- `assert_valid_wildcards` has been renamed to `dash_assert_valid_wildcards`
- the `dash` package now depends on `dashHtmlComponents`, `dashCoreComponents`, and `dashTable`
- the package help has additional details and links to helpful resources
- undocumented features are now documented in the `Dash` class

PR #99, opened 2019-06-25
Remove welcome_page() and layout_container_id()

rpkyle:
This PR proposes to :hocho: these Dash for R elements, which do not exist in Dash for Python:
- [x] `layout_container_id()`
- [x] `welcome_page()`
- [x] `is.layout()` -- this checked that `is.component()` returned `TRUE`, and that `id == layout_container_id()`

In addition, `componentify()` is renamed to `validate_component()`, which is somewhat clearer.

The primary motivation is to enforce parity with Dash for Python, and to properly expose the layout object to work with layout validation checks which are already in place.

@alexcjohnson

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -445,7 +445,7 @@ Dash <- R6::R6Class(
       if (render) private$layout_render() else private$layout_
     },
     layout = function(...) {
-      private$layout_ <- if (is.function(..1)) ..1 else list(...)
+      private$layout_ <- if (is.function(..1)) ..1 else (...) 
       # render the layout, and then return the rendered layout without printing
       invisible(private$layout_render())
     },
@@ -560,16 +560,8 @@ Dash <- R6::R6Class(
       # assuming private$layout is either a function or a list of components...
       layout_ <- if (is.function(private$layout_)) private$layout_() else private$layout_
 
-      # accomodate functions that return a single component
-      if (is.component(layout_)) layout_ <- list(layout_)
-
-      # make sure we are working with a list of components
-      layout_ <- lapply(layout_, private$componentify)
-
-      # Put the list of components into a container div. I'm pretty sure dash
-      # requires the layout to be one component, but R folks are used to
-      # being able to supply "components" to ...
-      layout_ <- dashHtmlComponents::htmlDiv(children = layout_, id = layout_container_id())
+      # ensure layout is a Dash component or collection of components
+      layout_ <- private$validate_component(layout_)
 
       # store the layout as a (flattened) vector form since we query the
       # vector names several times to verify ID naming (among other things)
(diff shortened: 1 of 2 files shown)


R/dash.R
```diff
@@ -731,7 +723,7 @@ Dash <- R6::R6Class(
                   other = other_files_map))
     },
 
-    componentify = function(x) {
+    validate_component = function(x) {
       if (is.component(x)) return(x)
       if (all(vapply(x, is.component, logical(1)))) return(x)
       stop("The layout must be a component or a collection of components", call. = FALSE)
```
alexcjohnson: In python the error we get is:
```
dash.exceptions.NoLayoutException:
Layout must be a dash component or a function that returns a dash component.
```

As discussed on slack earlier, seems like the same holds in R, we can't have a list/collection here either. And if this is where you'd see the error if layout was given a function with an invalid return, I'd give exactly the same error message as in Python.

That has knock-on effects - so `layout = function(...) {` should turn into `layout = function(value) {`, and this whole `validate_component` might as well be 🔪 since it'll just be `if (!is.component(x)) stop(...)`

R/dash.R
```diff
@@ -920,7 +912,7 @@ Dash <- R6::R6Class(
 # @param layout
 # @param component a component (should be a dependency)
 validate_dependency <- function(layout_, dependency) {
-  if (!is.layout(layout_)) stop("`layout` must be a dash layout object", call. = FALSE)
+  if (!is.component(layout_)) stop("`layout` must be a Dash component or collection of components", call. = FALSE)
```
alexcjohnson: Is there a case we can actually hit this `stop`? Doesn't the one in `layout_render` cover us?

rpkyle: I have trouble seeing where this gets tripped as well; this is leftover code from the early revisions of the package. I suspect it made more sense when we wrapped the layout in an `htmlDiv` and wanted to ensure the layout was a list. 

Here I'm actively trying to avoid that, so it probably does make sense to try :hocho: this.

alexcjohnson:
@rpkyle is this PR still relevant?

rpkyle:
> @rpkyle is this PR still relevant?

Yes, I believe so:

https://github.com/plotly/dashR/blob/fcfedcbbb6cd1cee71b782b66c62a7a1f18bcb54/R/utils.R#L18-L25

...and also:

https://github.com/plotly/dashR/blob/fcfedcbbb6cd1cee71b782b66c62a7a1f18bcb54/R/dash.R#L950-L951

But I think that ☝️ this is now redundant, and could be removed.

We did remove `welcome_page()` previously, however:

https://github.com/plotly/dashR/pull/96

alexcjohnson:
I notice this PR is also targeting a somewhat old branch, so perhaps it would be best to pluck the still-relevant pieces of this PR into a new branch, make a new PR, and close this one? Your call how to manage it, but it would be nice to wrap this up before it gets even more stale.

rpkyle:
I'll do that as soon as https://github.com/plotly/dashR/pull/111 is resolved.

rpkyle:
This PR has become stale, opening another which addresses the issues that remain and which are now summarized in #120.

PR #111, opened 2019-08-02
Provide support for no_update in Dash for R

### LINKED ISSUE #25: Implement exceptions.PreventUpdate
https://github.com/plotly/dash/pull/190

https://github.com/plotly/dash/commit/fd2f353814cbccc8fc9df378d780ebc96d7c136f
https://github.com/plotly/dash/commit/ad92fa31f0388813ba7695ef108504397c8c0ba2

rpkyle:
This PR proposes to introduce a function entitled `dashNoUpdate()` which functions analogously to `dash.no_update` in Python, although the implementation differs slightly.

As with its counterpart, the function permits developers to selectively prevent one or more outputs from updating the layout. Currently, `dashNoUpdate()` only supports single-output callbacks, but should be easily extended to the multiple outputs scenario when that feature is available in R. 

Additionally, the `response$body` returned is an empty character string, with a `response$status` of `204`:

https://github.com/plotly/dashR/blob/b23c24855fa8a7655464c38b907fba4c72c7bfbc/R/dash.R#L294-L299

Closes #25.

@Marc-Andre-Rivet

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -528,7 +528,7 @@ Dash <- R6::R6Class(
     # no_update may be used within a callback to prevent a single output
     # from updating; it returns a wrapped NULL of class "no_update"
     # ------------------------------------------------------------------------
-    no_update <- function() {
+    no_update = function() {
       x <- list(NULL)
       class(x) <- "no_update"
       return(x)
--- a/R/dash.R
+++ b/R/dash.R
@@ -287,11 +287,17 @@ Dash <- R6::R6Class(
         output_value <- getStackTrace(do.call(callback, callback_args),
                                       debug = private$debug,
                                       pruned_errors = private$pruned_errors)
-  
+        
         # reset callback context
         private$callback_context_ <- NULL
  
-        if (is.null(private$stack_message)) {
+        # inspect the output_value to determine whether any outputs have no_update
+        # objects within them; these should not be updated
+        if (class(output_value) == "no_update") {
+          response$body <- character(1) # return empty string
+          response$status <- 204L
+        }
+        else if (is.null(private$stack_message)) {
           # pass on output_value to encode_plotly in case there are dccGraph
           # components which include Plotly.js figures for which we'll need to 
           # run plotly_build from the plotly package
@@ -516,6 +522,16 @@ Dash <- R6::R6Class(
         warning("callback_context is undefined; callback_context may only be accessed within a callback.")
       }   
       private$callback_context_
+    },
+    
+    # ------------------------------------------------------------------------
+    # no_update may be used within a callback to prevent a single output
+    # from updating; it returns a wrapped NULL of class "no_update"
+    # ------------------------------------------------------------------------
+    no_update <- function() {
+      x <- list(NULL)
+      class(x) <- "no_update"
+      return(x)
     },
     
     # ------------------------------------------------------------------------
--- a/R/dash.R
+++ b/R/dash.R
@@ -293,7 +293,7 @@ Dash <- R6::R6Class(
  
         # inspect the output_value to determine whether any outputs have no_update
         # objects within them; these should not be updated
-        if (class(output_value) == "no_update") {
+        if (length(output_value) == 1 && class(output_value) == "no_update") {
           response$body <- character(1) # return empty string
           response$status <- 204L
         }
--- a/R/dash.R
+++ b/R/dash.R
@@ -287,7 +287,7 @@ Dash <- R6::R6Class(
         output_value <- getStackTrace(do.call(callback, callback_args),
                                       debug = private$debug,
                                       pruned_errors = private$pruned_errors)
-        
+
         # reset callback context
         private$callback_context_ <- NULL
  
--- a/R/dash.R
+++ b/R/dash.R
@@ -525,16 +525,6 @@ Dash <- R6::R6Class(
     },
     
     # ------------------------------------------------------------------------
-    # no_update may be used within a callback to prevent a single output
-    # from updating; it returns a wrapped NULL of class "no_update"
-    # ------------------------------------------------------------------------
-    no_update = function() {
-      x <- list(NULL)
-      class(x) <- "no_update"
-      return(x)
-    },
-    
-    # ------------------------------------------------------------------------
     # convenient fiery wrappers
     # ------------------------------------------------------------------------
     run_server = function(host = Sys.getenv('DASH_HOST', "127.0.0.1"), 
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -5,10 +5,14 @@
 #' Use in conjunction with the `callback()` method from the [dash::Dash] class
 #' to define the update logic in your application.
 #'
+#' The `dashNoUpdate()` function permits application developers to prevent a
+#' single output from updating the layout. It has no formal arguments, but
+#' returns an object of class `no_update`, for which Dash will return a `204`
+#' status code and empty response body.
+#' 
 #' @name dependencies
 #' @param id a component id
 #' @param property the component property to use
-
 
 #' @rdname dependencies
 #' @export
@@ -43,4 +47,12 @@ dependency <- function(id = NULL, property = NULL) {
     id = id,
     property = property
   )
+}
+
+#' @rdname dependencies
+#' @export
+dashNoUpdate <- function() {
+  x <- list(NULL)
+  class(x) <- "no_update"
+  return(x)
 }
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -6,9 +6,7 @@
 #' to define the update logic in your application.
 #'
 #' The `dashNoUpdate()` function permits application developers to prevent a
-#' single output from updating the layout. It has no formal arguments, but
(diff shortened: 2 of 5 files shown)


alexcjohnson:
Looks great!

Now that we have a test framework, let's get in the habit of testing new features as they come in here. You should be able to essentially just port this one test, but simplify it - take out the `Value` stuff, and you only have one variant to test as opposed to that test covering both `no_update` and `PreventUpdate`:

https://github.com/plotly/dash/blob/master/tests/integration/test_integration.py#L120-L178

R/dependencies.R
```diff
@@ -5,11 +5,15 @@
 #' Use in conjunction with the `callback()` method from the [dash::Dash] class
 #' to define the update logic in your application.
 #'
+#' The `dashNoUpdate()` function permits application developers to prevent a
+#' single output from updating the layout. It has no formal arguments, but
+#' returns an object of class `no_update`, for which Dash will return a `204`
+#' status code and empty response body.
```
alexcjohnson: users don't care about the implementation here - they just want to know that as a result the output prop will not change. I'd just drop everything past "It has no formal arguments"

rpkyle:
@alexcjohnson I identified a minor issue today while reviewing app PRs with the new code, the following warning appears periodically for some applications:

`warning: the condition has length > 1 and only the first element will be used from if (class(output_value) == "no_update") {`

I'll resolve this and comment in this PR when the commit is made, in addition to adding in the test we discussed.

rpkyle:
> Looks great!
> 
> Now that we have a test framework, let's get in the habit of testing new features as they come in here. You should be able to essentially just port this one test, but simplify it - take out the `Value` stuff, and you only have one variant to test as opposed to that test covering both `no_update` and `PreventUpdate`:
> 
> https://github.com/plotly/dash/blob/master/tests/integration/test_integration.py#L120-L178

fixed in d1f384f

@byronz

byronz:
> fixed in [d1f384f](https://github.com/plotly/dashR/commit/d1f384f1eeb6f7c208ae7988085dd0747c79529a)

@rpkyle   I would like a change in the test with the new proposed API in https://github.com/plotly/dash/pull/859

R/dash.R
```diff
@@ -287,11 +287,17 @@ Dash <- R6::R6Class(
         output_value <- getStackTrace(do.call(callback, callback_args),
                                       debug = private$debug,
                                       pruned_errors = private$pruned_errors)
-  
+
```
alexcjohnson: oops trailing whitespace...

tests/integration/callbacks/test_no_update.py
```diff
@@ -0,0 +1,63 @@
...
+             }
+)
+
+app$run_server()
+"""
+
+
+def test_rsnu001_no_update(dashr):
+    dashr.start_server(app)
+    dashr.find_element("#color-selector").click()
```
byronz: with the new API, you can do sth like this 
```suggestion
for color in ("#FF0000", "#00FF00", "#0000FF"):
    dashr.select_dcc_dropdown('#color-selector', value=color)
    assert  dashr.find_element('#message-box').text.endswith(color)
```

alexcjohnson:
The current test failure - where the value hasn't updated between the `.click()` call and the `assert` - is why I used `.wait_for_text_to_equal` in the dash-py test. In general there should be some sort of "wait for" after any callback, then additional tests can be bare `assert` statements.

In this case the `dashNoUpdate()` calls mean we'd be waiting for the text to be what it already was - so in fact if the test is fast enough we're not really testing anything. That's why the dash-py test uses the multiprocessing `Value` in `until(lambda: callback1_count.value == 4,...)`. Since we can't do that here, I'd suggest just waiting a little while (`time.sleep(1)`?) before testing the text when it's intended not to change.

tests/integration/callbacks/test_no_update.py
alexcjohnson: That looks good - but let's do that in a followup PR, so we can merge this before releasing the new dropdown API.

byronz:
> The current test failure - where the value hasn't updated between the `.click()` call and the `assert` - is why I used `.wait_for_text_to_equal` in the dash-py test. In general there should be some sort of "wait for" after any callback, then additional tests can be bare `assert` statements.
> 
> In this case the `dashNoUpdate()` calls mean we'd be waiting for the text to be what it already was - so in fact if the test is fast enough we're not really testing anything. That's why the dash-py test uses the multiprocessing `Value` in `until(lambda: callback1_count.value == 4,...)`. Since we can't do that here, I'd suggest just waiting a little while (`time.sleep(1)`?) before testing the text when it's intended not to change.

this represents a common scenario where you would like to check something after an action and would like to wait a little bit so the assertion was not made too early to test nothing. 

instead of the sleep then assert (our current way), I'm proposing to have sth similar to `wait.until` or `wait.until_not`: it does more assertions during that sleep period with a poll=0.1. 

```wait.check(lambda: out.text.endswith('#FF0000'), duration=0.5)```

**Comment by alexcjohnson** [comment]
@byronz not sure I quite follow you re: `wait.check` but let's have this discussion in an issue over in the dash repo. As in my previous comment I'd like to get this merged using the existing `dash.testing` API, then we can come back and clean it up in a future PR.

**Comment by byronz** [comment]
> @byronz not sure I quite follow you re: `wait.check` but let's have this discussion in an issue over in the dash repo. As in my previous comment I'd like to get this merged using the existing `dash.testing` API, then we can come back and clean it up in a future PR.

@alexcjohnson  I'm talking about change this right now in this PR, but something we might need as test API like `wait.check`, what it does is consistently check the condition for a specified amount of time.

**Comment by rpkyle** [comment]
> @alexcjohnson I identified a minor issue today while reviewing app PRs with the new code, the following warning appears periodically for some applications:
> 
> `warning: the condition has length > 1 and only the first element will be used from if (class(output_value) == "no_update") {`
> 
> I'll resolve this and comment in this PR when the commit is made, in addition to adding in the test we discussed.

fixed in ae7b1b0

File: R/dash.R (hunk 2 above)
Comment (rpkyle) [reply R/dash.R]: 🙈

Comment (rpkyle) [reply R/dash.R]: fixed in 01bc431b8f11711a6566f3d5c9b40054eade11b0

File: R/dash.R [hunk 4]
```diff
@@ -293,7 +293,7 @@ Dash <- R6::R6Class(
  
         # inspect the output_value to determine whether any outputs have no_update
         # objects within them; these should not be updated
-        if (class(output_value) == "no_update") {
+        if (length(output_value) == 1 && class(output_value) == "no_update") {
```
alexcjohnson: Great! What's the case where this extra condition is necessary? And which test exercises that condition? 😈

rpkyle: > Great! What's the case where this extra condition is necessary? And which test exercises that condition? 😈

:joy: I saw that one coming!

- the case is any `output_value` which is a vector or list with two or more elements (and for which `if` would return this warning since the function is not vectorized)
- I could add a button which returns a vector and then send a click to see if the warning doesn’t arise; is that sufficient?

alexcjohnson: > I could add a button which returns a vector and then send a click to see if the warning doesn’t arise; is that sufficient?

perfect.

R/dependencies.R
rpkyle: fixed in 10404e0

R/dash.R
rpkyle: fixed in 8f6dafd

rpkyle:
> The current test failure - where the value hasn't updated between the `.click()` call and the `assert` - is why I used `.wait_for_text_to_equal` in the dash-py test. In general there should be some sort of "wait for" after any callback, then additional tests can be bare `assert` statements.
> 
> In this case the `dashNoUpdate()` calls mean we'd be waiting for the text to be what it already was - so in fact if the test is fast enough we're not really testing anything. That's why the dash-py test uses the multiprocessing `Value` in `until(lambda: callback1_count.value == 4,...)`. Since we can't do that here, I'd suggest just waiting a little while (`time.sleep(1)`?) before testing the text when it's intended not to change.

fixed in 8f6dafd

tests/integration/callbacks/test_no_update.py
```diff
@@ -0,0 +1,103 @@
...
+    dashr.find_elements("div.VirtualizedSelectOption")[0].click()
+    time.sleep(1)
+    assert dashr.find_element("#message-box").text == "The hexadecimal representation of your last chosen color is #FF0000"
```
alexcjohnson: Not to beat a dead horse, but since this is one of the first tests I'd like to make sure we get the best pattern for all the other tests to build on.

`time.sleep` should be a last resort. In this test you only need it when the text is expected to be the same after the action as it was before - ie just the checks after `...[3].click()`. But in all the other cases you can do
```py
dashr.wait_for_text_to_equal(
    "#message-box",
    "The hexadecimal representation of your last chosen color is #FF0000"
)
```
Which will return as soon as the correct value is found on the page, saving us most of that 1 sec delay each time.

rpkyle: fixed in be9d387

PR #116, opened 2019-08-13
Better handling for user-defined error conditions in debug mode

### LINKED ISSUE #115: getStackTrace does not properly handle errors thrown by "stop"
Currently, `getStackTrace` supplies a pruned (concise, with non-Dash related calls removed) or raw (everything) traceback when a Dash app is initialized with `debug=TRUE`. 

This worked well for spontaneous errors, but did not properly handle user-defined `stop` conditions; the stack trace would not be displayed in the console or browser:

```
in debug mode, catching error as warning ...
warning: Execution error in stop: Winnipeg? from NULL
request: 127.0.0.1 - ID_127.0.0.1 [13/Aug/2019:10:34:57 -0400] "POST /_dash-update-component HTTP/1.1" 200 38 "http://127.0.0.1:8050/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_4) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.1 Safari/605.1.15"
```

In addition, in one instance, the following JS console error would be sent up to the renderer and appear in the dev tools UI:

Screenshot 2019-08-13 11 08 26

Addressing this issue could prove somewhat complicated, since `stop` calls are stored as [`symbol` objects](https://stat.ethz.ch/R-manual/R-devel/library/base/html/name.html), and can't be treated directly as strings:

```
Error in sprintf(template, header, throwing_call, error_message, formattedStack) : 
  invalid type of argument[2]: 'symbol'
```

A workaround might be to `deparse` the object, which will return a string. A solution should be identified such that `getStackTrace` properly handles user-defined errors, just as it does for more spontaneous exceptions.

rpkyle:
As noted in #115, `getStackTrace` behaves a bit unexpectedly when a user defined error is encountered within a callback while in debug mode, e.g.

```
    if (city == "Winnipeg")
      stop(simpleError("Winnipeg?"))
```

ought to produce the following console text while sending a traceback up to the dev tools UI:

```
 ### DashR Traceback (most recent/innermost call last) ###
    1: getStackTrace
    2: do.call
    3: <anonymous> function(city, tempscale)
    4: stop
in debug mode, catching error as warning ...
warning: Execution error in stop: Winnipeg? from NULL
```

However, the traceback was not provided previously in such cases, and a JS error could be thrown in the browser instead, halting app execution. 

This PR proposes to address the problem by introducing conditional logic to identify calls of class `simpleError`, and to deparse the calls so that they may be treated as strings when constructing the `stack_message` object. 

The problem is that `stop` is a `symbol` in R, and attempting to naively treat it as a string may throw the following error:

```
Error in sprintf(template, header, throwing_call, error_message, formattedStack) : 
  invalid type of argument[2]: 'symbol'
```

which can prevent the traceback from being returned. The key additions are

https://github.com/plotly/dashR/blob/b170cfa72dc52a335a3d86763fad731fa27fcb6c/R/utils.R#L698-L707

https://github.com/plotly/dashR/blob/b170cfa72dc52a335a3d86763fad731fa27fcb6c/R/utils.R#L738-L747

https://github.com/plotly/dashR/blob/b170cfa72dc52a335a3d86763fad731fa27fcb6c/R/utils.R#L762-L770

Closes #115.

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -286,7 +286,7 @@ Dash <- R6::R6Class(
 
         output_value <- getStackTrace(do.call(callback, callback_args),
                                       debug = private$debug,
-                                      pruned_errors = private$pruned_errors)
+                                      prune_errors = private$prune_errors)
   
         # reset callback context
         private$callback_context_ <- NULL
@@ -525,7 +525,7 @@ Dash <- R6::R6Class(
                           port = Sys.getenv('DASH_PORT', 8050), 
                           block = TRUE, 
                           showcase = FALSE, 
-                          pruned_errors = TRUE, 
+                          dev_tools_prune_errors = TRUE, 
                           debug = FALSE, 
                           dev_tools_ui = NULL,
                           dev_tools_props_check = NULL,
@@ -545,7 +545,7 @@ Dash <- R6::R6Class(
         self$config$props_check <- FALSE
       }
 
-      private$pruned_errors <- pruned_errors
+      private$prune_errors <- dev_tools_prune_errors
       private$debug <- debug
       
       self$server$ignite(block = block, showcase = showcase, ...)
@@ -569,7 +569,7 @@ Dash <- R6::R6Class(
     
     # initialize flags for debug mode and stack pruning,
     debug = NULL,
-    pruned_errors = NULL,
+    prune_errors = NULL,
     stack_message = NULL,
 
     # callback context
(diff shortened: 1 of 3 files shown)


rpkyle:
@alexcjohnson Any comments on this one?

rpkyle:
@alexcjohnson To test this properly, I think we will need to throw the error and compare with the expected string it produces within R. I believe it's important to ensure the error reaches the console, as noted in plotly/dashR/issues/115.

I'm not quite sure how to trap the R error using Selenium within Python, but I could try writing a `testthat` test which uses `RSelenium` to control the app instead.

Either way, I'd find it helpful to discuss further before I proceed.

rpkyle:
> As discussed this morning, I'll implement a basic test which incorporates a `stop` error to demonstrate what this resolves.

fixed in 1029ec48b4ee76c3c6b85ad999532876a98c5b38

PR #119, opened 2019-08-23
Provide support for multiple outputs

### LINKED ISSUE #114: Need support for multiple outputs
Dash for Python has supported multiple outputs since https://github.com/plotly/dash/pull/436
was merged, which resolved https://github.com/plotly/dash/issues/149.

Dash for R should add permit app developers to target multiple outputs via a single callback as well.

rpkyle:
This PR provides support for multiple outputs, as originally described in https://github.com/plotly/dash/pull/436. Several changes are proposed:
- a `createCallbackId` function is introduced to simplify generating output IDs for either single outputs (`container.style`) or multiple outputs (`..text-box.children...container.style..`)
- an `insertIntoCallbackMap` function is also introduced to add additional callback handlers into the map
- callback handlers may now return values which are targeted at multiple callbacks, and these targets are inferred solely from the sequence of the output declaration (i.e. no names are required for the return object elements)
- partial updates are properly supported
- modifications to CircleCI test config to install the correct package versions for a given `plotly/dashR` version, and to avoid repeatedly launching R for package installs

The following checks are now performed within Dash for R:
- `validateOutput` now checks the ID formatting for single and multiple outputs
- `assert_valid_callbacks` now checks for circular inputs and outputs for multiple outputs also
- `insertIntoCallbackMap` checks for duplicated outputs across callbacks before allowing a new callback to be added to the callback map

An additional [test](https://github.com/plotly/dashR/pull/119/commits/d77a0cec47c7e47e9ced64640295b066d76d24ff) is also proposed to ensure that multiple outputs are functioning properly, and that partial outputs are allowed.

Finally, the following checks are implemented as in Dash for Python:
- [x] Ensuring that no output is specified twice [`#f1f4724`](https://github.com/plotly/dashR/pull/119/commits/f1f472427172fdf4e100ef9560133922cd2601d7)
- [x] Ensuring that no output is used in two overlapping multi-output sets (e.g. you cannot have `..o1.value...o2.value..` with `..o1.value...o3.value..`) [`#508eeb1`](https://github.com/plotly/dashR/pull/119/commits/508eeb1bed50fa74553d679a625fb534a3a80b9b)
- [x] Ensuring that both methods of supplying an `output` parameter are working (e.g. using `output=list(id='graph', property='figure')` or `output=output(id='graph', property='figure')`) [`#5521a93`](https://github.com/plotly/dashR/pull/119/commits/5521a930107f78638b7b9673a30ffa1a0be87d8d)

Closes #114.

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -284,11 +284,11 @@ Dash <- R6::R6Class(
 
         output_value <- getStackTrace(do.call(callback, callback_args),
                                       debug = private$debug,
-                                      pruned_errors = private$pruned_errors)
+                                      prune_errors = private$prune_errors)
 
         # reset callback context
         private$callback_context_ <- NULL
-        
+ 
         # inspect the output_value to determine whether any outputs have no_update
         # objects within them; these should not be updated
         if (length(output_value) == 1 && class(output_value) == "no_update") {
@@ -561,7 +561,7 @@ Dash <- R6::R6Class(
                           port = Sys.getenv('DASH_PORT', 8050), 
                           block = TRUE, 
                           showcase = FALSE, 
-                          pruned_errors = TRUE, 
+                          dev_tools_prune_errors = TRUE, 
                           debug = FALSE, 
                           dev_tools_ui = NULL,
                           dev_tools_props_check = NULL,
@@ -581,7 +581,7 @@ Dash <- R6::R6Class(
         self$config$props_check <- FALSE
       }
 
-      private$pruned_errors <- pruned_errors
+      private$prune_errors <- dev_tools_prune_errors
       private$debug <- debug
       
       self$server$ignite(block = block, showcase = showcase, ...)
@@ -605,7 +605,7 @@ Dash <- R6::R6Class(
     
     # initialize flags for debug mode and stack pruning,
     debug = NULL,
-    pruned_errors = NULL,
+    prune_errors = NULL,
     stack_message = NULL,
 
     # callback context
--- a/R/dash.R
+++ b/R/dash.R
@@ -222,7 +222,7 @@ Dash <- R6::R6Class(
         payload <- Map(function(callback_signature) {
           list(
             inputs=callback_signature$inputs,
-            output=paste0(callback_signature$output, collapse="."),
+            output=createCallbackId(callback_signature$output),
             state=callback_signature$state
           )
         }, private$callback_map)
@@ -305,12 +305,38 @@ Dash <- R6::R6Class(
           
           # have to format the response body like this
           # https://github.com/plotly/dash/blob/064c811d/dash/dash.py#L562-L584
-          resp <- list(
-            response = list(
-              props = setNames(list(output_value), gsub( "(^.+)(\\.)", "", request$body$output))
+          if (grepl("\\.\\.\\.", request$body$output)) {
+            # if multi-output callback, isolate the output IDs and properties
+            idmatch <- gregexpr("(?<=\\.\\.)(([^\\.]+))(?=\\.)", request$body$output, perl=TRUE)
+            propmatch <- gregexpr("(?<=\\.)(([^\\.]+))(?=\\.\\.)", request$body$output, perl=TRUE)
+            ids <- unlist(regmatches(request$body$output, idmatch))
+            props <- unlist(regmatches(request$body$output, propmatch))
+            
+            # prepare a response object which has list elements corresponding to ids
+            # which themselves contain named list elements corresponding to props
+            # then fill in nested list elements based on output_value
+            
+            allprops <- setNames(vector("list", length(unique(ids))), unique(ids))
+            
+            idmap <- setNames(ids, props)
+            
+            for (id in unique(ids)) {
+              allprops[[id]] <- output_value[grep(id, ids)]
+              names(allprops[[id]]) <- names(idmap[which(idmap==id)])
+            }
+              
+            resp <- list(
+             response = allprops,
+             multi = TRUE
+             )
+          } else {
+            resp <- list(
+              response = list(
+                props = setNames(list(output_value), gsub( "(^.+)(\\.)", "", request$body$output))
+              )
             )
-          )
-          
+          }
+        
           response$body <- to_JSON(resp)
           response$status <- 200L
           response$type <- 'json'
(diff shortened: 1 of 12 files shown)


rpkyle:
A sample app is provided here:

```
library(dash)
library(dashHtmlComponents)
library(dashCoreComponents)
library(plotly)
library(dashTable)

sample_data <- list(
  series = list(
    data = list(
      list(title = 'Game of Thrones', 
           score = 9.5),
      list(title = 'Stranger Things', 
           score = 8.9),
      list(title = 'Vikings', 
           score = 8.6)
      ),
    style = list(
      backgroundColor = '#ff998a'
    )
  ),
  movies = list(
    data = list(
      list(title = 'Rambo', 
           score = 7.7),
      list(title = 'The Terminator', 
           score = 8.0),
      list(title = 'Alien', 
           score = 8.5)
      ),
    style = list(
      backgroundColor = '#fff289'
    )
  )
)

app <- Dash$new()       

app$layout(
  htmlDiv(list(
    htmlH1('Multi output example'),
    dccDropdown(id='data-dropdown',
                options = list(
                  list(label = 'Movies',
                       value = 'movies'),
                  list(label = 'Series',
                       value = 'series')
                ),
                value = 'movies'),
    htmlDiv(list(
      dccGraph(id = 'graph'),
      dashDataTable(id = 'data-table',
                    columns = list(
                      list(name = 'Title',
                           id = 'title'),
                      list(name = 'Score',
                           id = 'score')
                    )
      )
    )
    )
  ),
  id = 'container'
  )
)

app$callback(output=list(
  output(id='graph', property='figure'),
  output(id='data-table', property='data'),
  output(id='data-table', property='columns'),
  output(id='container', property='style')
  ),
  params=list(
    input(id='data-dropdown', property='value')
  ),
  function(value) {
    if (is.null(value)) {
      return(dashNoUpdate())
    }

    selected <- sample_data[[value]]
    
    subdata <- selected[['data']]

    columns <- list(
      list(
        name = 'Title',
        id = 'title'
      ),
      list(
        name = 'Score',
        id = 'score'
      )
    )
    
    figure <- plot_ly(
      x = vapply(subdata, function(x) x$score, numeric(1)),
      text = vapply(subdata, function(x) x$title, character(1)),
      name = vapply(subdata, function(x) x$title, character(1)),
      type = "bar"
    )
    
    browser()
    
    return(list(figure,
                subdata,
                columns,
                selected[['style']]))
})

app$run_server(debug=TRUE)
```

R/dash.R
```diff
@@ -305,12 +305,38 @@ Dash <- R6::R6Class(
...
           # https://github.com/plotly/dash/blob/064c811d/dash/dash.py#L562-L584
-          resp <- list(
-            response = list(
-              props = setNames(list(output_value), gsub( "(^.+)(\\.)", "", request$body$output))
+          if (grepl("\\.\\.\\.", request$body$output)) {
+            # if multi-output callback, isolate the output IDs and properties
+            idmatch <- gregexpr("(?<=\\.\\.)(([^\\.]+))(?=\\.)", request$body$output, perl=TRUE)
+            propmatch <- gregexpr("(?<=\\.)(([^\\.]+))(?=\\.\\.)", request$body$output, perl=TRUE)
+            ids <- unlist(regmatches(request$body$output, idmatch))
+            props <- unlist(regmatches(request$body$output, propmatch))
```
alexcjohnson: I'd generally try to avoid regex for simple things like this - my experience in Python and JS is it's slow and prone to errors if not heavily tested.

This could be something like:
```
if (substr(request$body$output, 1, 2) == '..') {
  outputs <- strsplit(substr(request$body$output,3,nchar(s)-2), '...', fixed=TRUE)
  id_props <- sapply(outputs, strsplit, '.', fixed=TRUE)
  ids <- lapply(id_props, '[', 1)
  props <- lapply(id_props, '[', 2)
```

R/dash.R
```diff
@@ -305,12 +305,38 @@ Dash <- R6::R6Class(
           
           # have to format the response body like this
           # https://github.com/plotly/dash/blob/064c811d/dash/dash.py#L562-L584
```
alexcjohnson: whew that's an old link! Perhaps update it, split the comment to cover single and multi separately?

Also I notice up a few lines is the 204 response for `no_update` - it's probably not strictly necessary but the py side turns multi-outputs where *all* outputs have been canceled into 204's as well.

R/dash.R
rpkyle: Both `grep` and `grepl` are part of base R and written as internal functions using native C code. I don't think performance is an issue here (is it possible to give an example or sample code?), but I'll rewrite this block if it will help avoid errors and unintended behaviour.

rpkyle:
Finishing up tests for the second and third checked items above, but I believe the current version of this code satisfies all requirements noted above.

rpkyle:
@byronz I made some minor changes to the CircleCI config in 30ba139497c18c89c3d60f432a72bb1174127145:
- R package dependencies are now installed on one line, so we don't repeatedly launch R to complete all package installs before the tests start
- we install `plotly/dashR` instead of the core packages, since the latter approach causes us to grab versions that are not locked to the current Dash for R release

rpkyle:
@alexcjohnson I've implemented the unit and integration tests for all features/scenarios requested, except for the most recent one you asked about, which I'll do today.

rpkyle:
@alexcjohnson OK, test has been added with two `DashNoUpdate()` returns in the same callback.

tests/integration/callbacks/multiple_outputs.py
```diff
@@ -0,0 +1,112 @@
...
+  output(id='text-box', property='children'),
+  output(id='container', property='style')
+),
+params=list(
+  input(id='data-dropdown', property='value'),
+  input(id='radio-partial', property='value')
+),
+function(value, choice) {
+  if (is.null(value)) {
+    return(dashNoUpdate())
```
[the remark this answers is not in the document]
rpkyle: fixed in 029af1c

tests/integration/callbacks/test_no_update_multiple.py
```diff
@@ -0,0 +1,113 @@
...
+    dashr.wait_for_text_to_equal(
+        "#output-state",
+        'Input 1 is Quebec, and Input 2 is Canada'
+    )
+    dashr.wait_for_text_to_equal(
+        "#output-clicks",
+        'The Button has been pressed 2 times.'
+    )
+    dashr.find_elements("input[type='checkbox']")[1].click()
+    dashr.find_element("#submit-button").click()
```
alexcjohnson: This is a great test 🎉 only thing I'd suggest, to make it easier to understand: before each `("#submit-button").click()` (or whatever it is that triggers the callback) add a comment about which checkboxes are selected, like:
`# Neither checkbox is selected, so neither output should update`

rpkyle: fixed in 8071866

PR #121, opened 2019-08-23
Remove is.layout, layout_container_id, and redundant component test

### LINKED ISSUE #120: Remove htmlDiv layout container, is.layout()
This issue describes changes to bring the Dash for R internal API more in line with the Python version of the backend. The changes here are minimal, and related largely to outdated code:
- :hocho: `is.layout()` and `layout_container_id()`
- :hocho: redundant `is.component(layout_)` [test](https://github.com/plotly/dashR/pull/99/commits/3a38d40a05fe56ec5bf703bad0bd5b3e11cdc82e#r297435510)

https://github.com/plotly/dashR/blob/9aace061a7a957fc79285cc21a87a212489ecf1b/R/utils.R#L18-L25

rpkyle:
This PR proposes changes to Dash for R layout handling, in efforts to bring this aspect of the internal API in line with Dash for Python:
- :hocho: `is.layout()` and `layout_container_id()`
- :hocho: redundant `is.component(layout_)` [test](https://github.com/plotly/dashR/pull/99/commits/3a38d40a05fe56ec5bf703bad0bd5b3e11cdc82e#r297435510)
- :hocho: wrapping `layout_` within a `list` step within `layout_render`
- remove implicit dependency on `dashHtmlComponents`

https://github.com/plotly/dashR/blob/9aace061a7a957fc79285cc21a87a212489ecf1b/R/utils.R#L18-L25

The original implementation of `layout_render` sought to allow users to pass components as `...`, which is non-standard for Dash. We now use `...` for wildcards within components (such as `html-*` and `aria-*`), so this is no longer possible.

Closes #120.


rpkyle:
This PR should fix the issue noted in https://github.com/facultyai/dash-bootstrap-components/issues/370; I'll test that package when Dash for R is rebuilt and post an update with my findings.

rpkyle:
@alexcjohnson @Marc-Andre-Rivet I believe I have addressed the underlying issue which prevented merging this PR previously. The problem was related to the original implemention of `layout()`, which had remained essentially unchanged since the early development of Dash for R:

https://github.com/plotly/dashR/blob/9c5a30760e3be985c866f122e9c2bd6ce2a516a6/R/dash.R#L774-L778

The `...` is vaguely analogous to `kwargs` in Python, and so the earlier code was permitting layouts in which multiple arguments could be used to pass components or functions that return components, e.g.

```r
app$layout(
   htmlDiv("a", id="a"),
   htmlDiv("b", id="b")
)
```

was seen as valid. This made excising the call to `htmlDiv` which occurred below difficult, since much of the original `layout_render()` code was intended to enable users to pass a function, a component with zero or more `children`, _or_  a series of components as arguments.

This PR now implements what I believe to be the correct behaviour, and tests to ensure that the syntax above will fail to update the layout. This would be a breaking change in the official release, so I think it's good to tie up this very loose end prior to releasing on CRAN.

rpkyle:
> 💃 Looks great, nice to have this match the Python side more closely.

Yeah, not sure how I didn't pick up on this before, but the immediate impact is that users of third-party components (such as `dashBootstrapComponents`) won't always need to load `dashHtmlComponents` when building apps when it isn't explicitly required.

> 
> Perhaps at some point we would want the renderer itself to support a list of components as the top-level layout - and then we could allow that in the back ends too. But now is not that time 😁

That seems like it could be potentially desirable, if it doesn't cause sibship issues within React. Seems like `<React.Fragment>` could help with that:
https://medium.com/@jmegilman/react-error-adjacent-jsx-elements-must-be-wrapped-in-an-enclosing-tag-32bc21c4434b

Anyway, glad to see this one finally getting put to 🛌 .

PR #127, opened 2019-09-18
Provide support for hot reloading in Dash for R

### LINKED ISSUE #125: Need to implement hot reloading support in Dash for R
Dash for Python offers a lovely "hot-reloading" feature, as described in https://github.com/plotly/dash/issues/66, https://github.com/plotly/dash/pull/362, and https://github.com/plotly/dash-renderer/pull/73. Supporting this as in the Python implementation would require
- creating a hash at app initialization, and regenerating it on file change
- adding a route to serve the hash
- front-end hot reloading
- back-end hot reloading (reload the app by restarting the `fiery` session)
- adding parameters for the dev tools UI: `dev_tools_hot_reload`, `dev_tools_hot_reload_interval`, `dev_tools_hot_reload_watch_interval`, `dev_tools_silence_routes_logging`

On the Python side, we use the Flask debug reloader to reset the hash. `fiery` doesn't have native support for something like this feature, but it does allow "Custom Events" which can be manipulated to serve a similar purpose.

State-preserving hot reloading would be really nice to have, as noted in https://github.com/plotly/dash/issues/460, but it has not been implemented in Dash for Python yet either.

@alexcjohnson

rpkyle:
This PR proposes to support frontend and backend hot reloading in Dash for R, in a manner analogous to that currently provided in Dash for Python (plotly/dash#66, plotly/dash#362, plotly/dash-renderer#73).

The primary modifications required include
- [x] Identifying the local path to the running Dash app, and storing this within the `Dash` object
- [x] Tracking file changes, both within the specified `assets` folder, but also within the app directory root itself
- [x] Generating a reload hash and a route to supply it on demand to the frontend
- [x] Formatting a JSON response to the frontend, including a modified file manifest
- [x] Handle partial reloading for CSS modifications only as in Dash for Python
- [x] Providing a custom Fiery logger which silences routing and request traffic on demand using the `dev_tools_silence_routes_logging` parameter, as in Dash for Python
- [x] Providing support for restarting the Fiery server when `hard_reload` is set (e.g. if non-CSS assets or the app itself is modified), and running `source` to reload the app code

:sparkles: It also proposes a new (logical) `viewer` parameter, which can be used to load Dash apps within RStudio for a more pleasant development experience. This feature requires specifying `127.0.0.1` or `localhost` for the `host` parameter; this is a restriction imposed by RStudio's API.

https://www.rdocumentation.org/packages/rstudioapi/versions/0.10/topics/viewer

Additional work is required:
- [x] To format and return a JSON response on file modification, and to ensure that reloading for modifications to CSS only behaves as in Dash for Python (soft reload, style should change without losing state)
- [x] I'll also need to consider implementation of `dev_tools_hot_reload_watch_interval`; this is used to control how frequently `walk_assets_directory` is invoked.
- [x] 🚨 We'll need a minimal test suite to verify that this feature behaves as desired.
- [x] Add in relevant R help content for (internal, package-level) documentation
- [x] Update `CHANGELOG.md` for `dev` branch

Closes #125.

@alexcjohnson @StalkComrade


rpkyle:
@alexcjohnson It took me quite some time to puzzle through a revised approach for this one, since @CanerIrfanoglu reported stack overflow and performance issues with the previous (poor) implementation.

Briefly, this version (still in progress, but functional) does the following:
- :hocho: `getwd()` and `app.R` references; user must load app via `source()` or `Rscript` for automatic reloading when the document containing the app code is modified
- :hocho: the `while` loop
- when "hot" reloading, a temporary environment is created, into which the objects referenced in the sourced document are loaded
- a `dash_nested_fiery_server` variable is created in the temporary environment
- `run_server` checks on startup for the presence of `dash_nested_fiery_server`; if it exists, `run_server` aborts gracefully
- the `app` object within the temporary environment is introspected for its `callback_map` and `layout_`, and these replace the contents of these fields within the currently running `app` object
- the temporary environment is removed and the event cycle begins again

R/dash.R
```diff
@@ -703,7 +737,7 @@ Dash <- R6::R6Class(
             # if any filetypes other than CSS are encountered in those which
             # are modified or deleted, restart the server
             
-            hard_reload <- updated_root || (has_assets && other_changed || other_added)
+            hard_reload <- updated_root || (has_assets && other_changed || other_added || other_deleted)
```
rpkyle: The implementation has been modified such that deleting a file (which is not `*.css`) triggers a full page reload also.

rpkyle:
@alexcjohnson I think the revised version is ready for initial review; still have two test failures, will try to sort out why they won't run in CircleCI this weekend.

rpkyle:
> * [ ]  Test a hard-reload scenario - same as above but the callback output should reset. I guess that would mean putting the app itself into a file and editing that file similarly to the css edit...

I think we actually need this test on the Python side also 🙂

R/dash.R
```diff
@@ -487,13 +488,9 @@ Dash <- R6::R6Class(
       dash_reload_hash <- paste0(self$config$routes_pathname_prefix, "_reload-hash")
       route$add_handler("get", dash_reload_hash, function(request, response, keys, ...) {
         modified_files <- private$modified_since_reload
-        hard <- FALSE
+        hard <- TRUE
```
rpkyle: @alexcjohnson once we 🔪 `hard` from `dash-renderer`, as outlined in plotly/dash#977, I'll 🔪 it here too.

rpkyle:
> * [x]  Test that CSS changes do a soft reload. The clearest way I can think of to do this is
>   
>   * make an app with a callback
>   * trigger that callback (eg by typing in an input box) and verify it shows up on the page
>   * change the CSS and verify that it shows up on the page
>   * check that the callback output has not been reset

fixed in 3f0b6ae25c420e60d8a04eb9e6aea8509d51136a

rpkyle:
> * [x]  Test a hard-reload scenario - same as above but the callback output should reset. I guess that would mean putting the app itself into a file and editing that file similarly to the css edit...

fixed in 99793467dea6c7002d0ebe0c24cf2e01b951e822

rpkyle:
> * [x]  Update `CHANGELOG.md` for `dev` branch

fixed in da30cb2b5aa2dadc00a91118d08a1c1715140135

CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
...
+- Hot reloading now supported in debug mode [#127](https://github.com/plotly/dashR/pull/127)
+- Support for displaying Dash for R applications within RStudio's viewer pane when `use_viewer = TRUE`
+- Clientside callbacks written in JavaScript are now supported [#130](https://github.com/plotly/dashR/pull/130)
+- Multiple outputs are now supported [#119](https://github.com/plotly/dashR/pull/119)
+- Selective callback updates to properties now supported with `dashNoUpdate()` [#111](https://github.com/plotly/dashR/pull/111)
+- Additional line number context inserted when available within stack traces [#133](https://github.com/plotly/dashR/pull/133)
+- Integration and unit tests are now performed when commits are made to open pull requests
+
+### Changed
+- Dash for R no longer requires forked `reqres`, patch applied upstream [#9](https://github.com/thomasp85/reqres/pull/9)
```
alexcjohnson: 🐄 A bare `#9` would imply this repo
This is how GH formats cross-repo links in comments:
```suggestion
- Dash for R no longer requires forked `reqres`, patch applied upstream [thomasp85/reqres#9](https://github.com/thomasp85/reqres/pull/9)
```

rpkyle:
> * [x]  Add in relevant R help content for (internal, package-level) documentation

fixed in 79a4996

rpkyle:
@alexcjohnson I think this one is ready to go 🚢 . I also added in documentation for `clientsideFunction`, which was previously missing.

R/dash.R
```diff
@@ -498,11 +508,55 @@ Dash <- R6::R6Class(
...
+        response$type <- 'json'
+        # reset the field for the next reloading operation
+        private$modified_since_reload <- list()
+        TRUE
+      })
+      
       router$add_route(route, "dashR-endpoints")
       server$attach(router)
-
+
```
alexcjohnson: 🙄

R/dash.R
```diff
@@ -572,33 +626,188 @@ Dash <- R6::R6Class(
...
+        } else {
+          return(default)
+        }
+      }
+      
+      self$server$host <- getServerParam(host, "character", "127.0.0.1")
+      self$server$port <- getServerParam(as.integer(port), "numeric", 8050)
+      
+      dev_tools_ui <- getServerParam(dev_tools_ui, "logical", NULL)
+      dev_tools_props_check <- getServerParam(dev_tools_props_check, "logical", NULL)
```
alexcjohnson: Do you need an `as.logical` here (and all the `"logical"` items), for the `getenv` case? In fact, perhaps `getServerParam` could be written to do this coercion, at least if you start with a string? 

This looks like a good solution to me - one kind of funny thing is users can provide all the `run_server` args as strings - `app$run_server(port="1234")` as well as `app$run_server(port=1234)` - but a nice side effect is we won't accept `run_server` args that don't end up with the correct type.

BTW it would be cleaner to set `debug` first, then use that as the default for some of these other args - rather than needing the confusing `is.null(x) && debug || isTRUE(x)` clauses.

R/dash.R
rpkyle: fixed in aa1ad07c4e5154ee96688b0158c3a1b22d9c7e0b

R/dash.R
rpkyle: Good point. I'll make that change, you're right that the handling for logicals-as-strings is missing this critical detail.

As for the confusing clauses, I think we'll still need a statement like

```
if (debug || isTRUE(dev_tools_ui))
  self$config$ui <- TRUE
else
  self$config$ui <- FALSE
```

because you cannot evaluate a `NULL` as a logical, e.g.

```
foo <- NULL
if (foo)
  cat("worked!\n")
```

will fail with `Error in if (foo) cat("worked!\n") : argument is of length zero`.

rpkyle: fixed in 6131677607c86a11e9760c68d1afd61dda4f64c6

R/dash.R
```diff
@@ -573,32 +627,178 @@ Dash <- R6::R6Class(
...
-      self$server$port <- as.numeric(port)
+      if (exists("dash_nested_fiery_server", env=parent.frame(1))) {
+        # fiery is attempting to launch a server within a server, abort gracefully
+        return(NULL)
+      }
 
-      if (is.null(dev_tools_ui) && debug || isTRUE(dev_tools_ui)) {
-        self$config$ui <- TRUE
+      getServerParam <- function(value, type, default) {
+        if (!is.null(value) && toupper(value) %in% c("TRUE", "FALSE"))
```
alexcjohnson:
```suggestion
        if (type == "logical" && !is.null(value) && toupper(value) %in% c("TRUE", "FALSE"))
```
Maybe also add a clause
```
if (type == "numeric")
    value <- as.numeric(value)
```
for the hot reload intervals; `as.integer` probably has to stay where it is since that doesn't appear to map onto a type...
```

Comment (rpkyle) [reply R/dash.R]: R is interesting in that there is no "type" function, but you can verify that an integer is of mode `numeric` and storage mode `integer`:

```
> foo <- 1L
> mode(foo)
[1] "numeric"
> storage.mode(foo)
[1] "integer"
```

The `col.info` function I wrote a while back attempts to present a summary with all the type/mode information I generally want for vectors and data.frames:

https://github.com/rpkyle/cscmisc/blob/master/R/col_info.R

rpkyle: fixed in ba7c01589f3cdd5d084a3a6b7f1ac733cb4d6458

rpkyle:
@alexcjohnson As a follow-up to our offline conversation, I've tried to make handling of (what ought to be) numeric dev tools parameters a bit more robust in 2e8b90c9655e3cb6561fe140039d62d35a6c6d1e.

PR #130, opened 2019-09-22
Implement support for clientside callbacks in Dash for R

### LINKED ISSUE #129: Need support for clientside callbacks
Dash for Python provides an interface for describing data transformation relationships between components in JavaScript (plotly/dash/issues/266, plotly/dash/pull/672, plotly/dash-renderer/pull/143). 

Updates which happen via this mechanism are triggered entirely client-side, without passing information over the network. Significant performance gains are possible relative to `@app.callback` functions in Python or `app$callback` functions in R.

Dash for R should offer support for so-called "clientside callbacks", just as Dash for Python does.

rpkyle:
This PR proposes modifications to support clientside callbacks in Dash for R, as they currently exist in Dash for Python (plotly/dash/issues/266, plotly/dash/pull/672, plotly/dash-renderer/pull/143).

Specifically, the following changes are proposed:
- add a function called `clientsideFunction` which allows app developers to reference a JavaScript function and its namespace
- modify `insertIntoCallbackMap` to accept either an R function or a JavaScript function
- modify `assert_valid_callbacks` to permit either a function to be passed into the `callback_map`, or a reference to a JavaScript function (provided that the object's names are `namespace` and `function_name`)
- modify the `_dash-dependencies` route handler to add `clientside_function` to the JSON payload

Closes #129.


rpkyle:
Sample app:

```
library(dash)
library(dashCoreComponents)
library(dashHtmlComponents)

app <- Dash$new()

app$layout(htmlDiv(list(
    dccInput(id='input'),
    htmlDiv(id='output-clientside'),
    htmlDiv(id='output-serverside')
    )
    )
)

app$callback( 
  output(id = "output-serverside", property = "children"),
  params = list(
    input(id = "input", property = "value")
  ),
  function(value) {
    sprintf("Server says %s", value)
  }
)

app$clientside_callback(
  output('output-clientside', 'children'),
  params=list(input('input', 'value')),
  clientsideFunction(
    namespace = 'clientside',
    function_name = 'display'
    )
)

app$run_server(showcase = TRUE, debug = TRUE)
```

tests/integration/callbacks/test_clientside.py
```diff
@@ -0,0 +1,72 @@
...
+)
+
+app$clientside_callback(
+  output('output-clientside', 'children'),
+  params=list(input('input', 'value')),
+  clientsideFunction(
+    namespace = 'clientside',
+    function_name = 'display'
+    )
+)
```
alexcjohnson: What if we allow `app$callback` to accept either an R `function` or a `clientsideFunction`, determine which it is based on type, and do away with `app$clientside_callback`? It's a little different from Python, but that's just because of Python's decorators vs R's anonymous functions. It's still plenty clear, I think, as you've still got an explicit call to `clientsideFunction`.

tests/integration/clientside/test_clientside.py
```diff
@@ -0,0 +1,73 @@
+from selenium.webdriver.support.select import Select
+import time, os
```
alexcjohnson: duplicate test files? Also may be nice to include in the test a second clientside function that takes 2 inputs.

rpkyle: Yeah, was in the middle of :hocho: these, sorry I missed this one earlier!

rpkyle: fixed in b85c6b7

tests/integration/callbacks/test_clientside.py
rpkyle: I wondered about that as I was working on this PR; using the same syntax for both seems perfectly fine to me. I'll go ahead and make this change.

rpkyle: fixed in e4fffaf

R/utils.R
```diff
@@ -413,7 +414,9 @@ assert_valid_callbacks <- function(output, params, func) {
 
   # Assert that user_function is a valid function
   if(!(is.function(func))) {
-    stop(sprintf("The callback method's 'func' parameter requires a function as its argument. Please verify that 'func' is a valid, executable R function."), call. = FALSE)
+    if (!(all(names(func) == c("namespace", "function_name")))) {
+      stop(sprintf("The callback method's 'func' parameter requires an R function or clientside_function call as its argument. Please verify that 'func' is either a valid R function or clientside_function."), call. = FALSE)
```
alexcjohnson:
```suggestion
      stop(sprintf("The callback method's 'func' parameter requires an R function or clientsideFunction call as its argument. Please verify that 'func' is either a valid R function or clientsideFunction."), call. = FALSE)
```

rpkyle:
> Looks great! just one small suggestion https://github.com/plotly/dashR/pull/130/files#r330202703 and trailing whitespace, then 💃

Right, sorry about that. RStudio has been configured to :hocho: trailing whitespace, but my `vim` settings weren't working as effectively. I had configured

`autocmd FileType R,Rd,py,js autocmd BufWritePre <buffer> %s/\s\+$//e`

Using a keystroke to make the change manually before save is working, so will use that going forward:

`nnoremap <F5> :let _s=@/<Bar>:%s/\s\+$//e<Bar>:let @/=_s<Bar><CR>`

PR #133, opened 2019-10-16
Add line number context to stack traces when srcrefs are available

### LINKED ISSUE #132: Need support for including line number references in stack traces when available
Since #87, we have provided stack trace messages when `debug=TRUE`:

image

These messages have lacked line number information, which limits their utility for debugging purposes since the location of the code which resulted in the error is not provided. In #127, the path to the source file is now stored whenever `source` or `Rscript` are used to launch the app.

This has the side benefit of enabling retrieval of line numbers via the `srcref` attribute, using `srcfile`:

https://stat.ethz.ch/R-manual/R-devel/library/base/html/srcfile.html

We should take advantage of this functionality to include line numbers when available.

rpkyle:
This PR proposes to inspect `srcref` attributes, when available, for additional context when `debug=TRUE` and errors are reported in the console or the dev tools UI in-browser. This information is populated in several settings:
- for internal, packaged functions, when `KeepSource` is `true` in the package `DESCRIPTION` file
- when applications are interactively loaded via `source()` and `keep.source = TRUE` (the default in most cases)
- when `Rscript` is used to invoke `source` at the CLI, and `keep.source = TRUE` is explicitly provided as an argument to `source` (since `keep.source` does not default to `TRUE` in most cases)

Some function calls will not have `srcref` attributes populated, and for these no line number is currently displayed. We will continue to investigate whether this functionality can be provided.

The initially proposed implementation looks like this:

image

... and within the dev tools UI, it looks like this:

image

Since there are enclosing environments between the error handler within `getStackTrace` and the `private` field to which `stack_message` is passed, it is necessary to resolve the number of frames to ascend up the stack. The proposed `countEnclosingFrames` function is an initial attempt at introducing this functionality.

This PR also resolves an issue with the option to display apps within RStudio's viewer pane, and renames that option from `viewer` to `use_viewer`.

Closes #132. @chriddyp

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -696,7 +696,7 @@ printCallStack <- function(call_stack, header=TRUE) {
         ": ",
         call_stack,
         " ",
-        lapply(call_stack, attr, "lineref")
+        lapply(call_stack, attr, "flineref")
         )
     ),
     stderr()
@@ -775,7 +775,9 @@ getStackTrace <- function(expr, debug = FALSE, prune_errors = TRUE) {
               currentCall <- deparse(completeCall)[1]
             else
               currentCall <- completeCall[[1]]
-            attr(currentCall, "lineref") <- getLineWithError(completeCall, formatted=TRUE)
+            
+            attr(currentCall, "flineref") <- getLineWithError(completeCall, formatted=TRUE)
+            attr(currentCall, "lineref") <- getLineWithError(completeCall, formatted=FALSE)
 
             if (is.function(currentCall) & !is.primitive(currentCall)) {
               constructedCall <- paste0("<anonymous> function(",
--- a/R/utils.R
+++ b/R/utils.R
@@ -684,7 +684,7 @@ encode_plotly <- function(layout_objs) {
 # so that it is pretty printed to stderr()
 printCallStack <- function(call_stack, header=TRUE) {
   if (header) {
-    write(crayon::yellow$bold(" ### DashR Traceback (most recent/innermost call last) ###"), stderr())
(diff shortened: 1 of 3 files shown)


R/utils.R
```diff
@@ -773,7 +789,7 @@ getStackTrace <- function(expr, debug = FALSE, prune_errors = TRUE) {
             }
 
           })
-
+
```
alexcjohnson: ?

R/utils.R
```diff
@@ -836,8 +850,24 @@ getStackTrace <- function(expr, debug = FALSE, prune_errors = TRUE) {
...
+getLineWithError <- function(currentCall, formatted=TRUE) {
+  srcref <- attr(currentCall, "srcref", exact = TRUE)
+  if (!is.null(srcref) & !(getAppPath()==FALSE)) {
+    # filename
+    srcfile <- attr(srcref, "srcfile", exact = TRUE)
+    # line number
+    if (formatted) {
+      crayon::yellow$italic(sprintf('Line %s in %s', srcref[[1]], srcfile$filename))
+    } else {
+      sprintf('Line %s in %s', srcref[[1]], srcfile$filename)
```
alexcjohnson: I'd reverse the order, and to make it easier to scan particularly in the unformatted case add a separator... something like `sprintf('-- %s, Line %s', srcfile$filename, srcref[[1]])`

Also 🌴 should only need `sprintf` once

R/utils.R
rpkyle: fixed in d4d4f2e

R/utils.R
rpkyle: fixed in ff632ce

PR #136, opened 2019-11-01
Disable percy finalize in test fixtures and use pytest --headless

rpkyle:
This PR proposes to make the following changes to the CircleCI configuration:

L40 is changed from
```
pytest tests/integration/
```

to

```
pytest tests/integration/ --headless --nopercyfinalize
```

and L41 is added:

```
percy finalize --all
```

which follows the guidance given in the Percy Snapshots section [here](https://dash.plot.ly/testing).

@alexcjohnson

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,8 @@ jobs:
             git clone --depth 1 https://github.com/plotly/dash.git
             cd dash && pip install -e .[testing] --quiet && cd ..
             export PATH=$PATH:/home/circleci/.local/bin/
-            pytest tests/integration/
+            pytest tests/integration/ --headless --nopercyfinalize
+            percy finalize --all
 
       - run:
           name: 🔎 Unit tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
             git clone --depth 1 https://github.com/plotly/dash.git
             cd dash && pip install -e .[testing] --quiet && cd ..
             export PATH=$PATH:/home/circleci/.local/bin/
-            pytest tests/integration/ --headless --nopercyfinalize
+            pytest --headless --nopercyfinalize tests/integration/
             percy finalize --all
 
       - run:


.circleci/config.yml
```diff
@@ -37,7 +37,8 @@ jobs:
             git clone --depth 1 https://github.com/plotly/dash.git
             cd dash && pip install -e .[testing] --quiet && cd ..
             export PATH=$PATH:/home/circleci/.local/bin/
-            pytest tests/integration/
+            pytest tests/integration/ --headless --nopercyfinalize
```
alexcjohnson:
```suggestion
            pytest --headless --nopercyfinalize tests/integration/
```
not entirely sure if this matters, but normally the flags come before the other args.

alexcjohnson:
@byronz using `--nopercyfinalize` didn't prevent us from *initializing* many percy runs. Apparently one per `test_*.py` file - this branch has 6 and created 6 runs, the hot reloading has 8 and created 8 runs.

I'm tempted to [try adding](https://github.com/plotly/dash/blob/b3b9de9923d82ff1281889b3695d638e95bb1590/dash/testing/browser.py#L78) something like
```
if not self.percy_runner._current_build:
    self.percy_runner.initialize_build()
```
([ref](https://github.com/percy/python-percy-client/blob/b14549998065e539446728f9c92a72e2993ce3a7/percy/runner.py#L67))

But I wonder why this hasn't hit us in dash core? We have multiple files contributing tests within a single container (as well as across containers), and there we collect everything in a single run. What's going wrong in this case?

byronz:
You also need to configure the Percy settings like we did in dash core. So it has different way to determinate build

.circleci/config.yml
```diff
@@ -37,7 +37,8 @@ jobs:
             git clone --depth 1 https://github.com/plotly/dash.git
             cd dash && pip install -e .[testing] --quiet && cd ..
             export PATH=$PATH:/home/circleci/.local/bin/
-            pytest tests/integration/
+            pytest --headless --nopercyfinalize tests/integration/
+            percy finalize --all
```
byronz: this `percy finalize --all` will not work, it needs to be a separate job

alexcjohnson:
Alright, @byronz we're going to just disable percy in this repo for now, will leave it up to you to figure out a sane way to manage this - but no rush :)

Issue #135, opened 2019-10-31
Inline clientside callbacks in Dash for R


rpkyle:
As noted in plotly/dash#956 and plotly/dash#967, it would be very convenient for app developers who wish to leverage JavaScript callbacks to compose these directly within the callback method. One approach might be to pass the JS directly as a string:

```
app$callback(
  output('out-component', 'value'),
  params=list(input('in-component1', 'value'), input('in-component2', 'value')),
  clientsideFunction(
    "
    function(largeValue1, largeValue2) {
        return someTransform(largeValue1, largeValue2);
    }
    "
    )
)
```

The namespace / function name can be determined in a fashion similar as what's done in DashPy (https://github.com/plotly/dash/pull/967), or not, with the `_dashprivate_` prefix.

This feature should be reasonably straightforward to implement in Dash for R, and it would help improve developer ergonomics if it were available.

rpkyle:
Resolved by #140; closing

PR #140, opened 2019-11-04
Support inline clientside callbacks in Dash for R

rpkyle:
This PR proposes to allow passing JavaScript-functions-as-strings within callbacks in Dash for R, addressing #135. This implementation fixes the `namespace` as `_dashprivate_`, and uses `output$id` for the function name, as in Python.

To do:
- [x] Create simple integration test to demonstrate the functionality

@Marc-Andre-Rivet @HammadTheOne


R/dash.R
```diff
@@ -593,17 +593,47 @@ Dash <- R6::R6Class(
...
+        # register the function with its hashed name representation
+        js_map <- insertIntoJSMap(private$js_map,
+                                  func,
+                                  fn_name)
+        
+        # check that no entries in the inline JS map have names matching those in the
+        # callback map, and vice-versas
+        if (length(intersect(names(private$callback_map), 
+                             sub("_dashprivate_clientside_", "", names(private$js_map)))) > 0) {
+          stop(sprintf("One or more outputs are duplicated across callbacks. Please ensure that all ID and property combinations are unique."), call. = FALSE)
```
alexcjohnson: Isn't this taken care of by `insertIntoCallbackMap` (that this still goes through below)?

R/utils.R
```diff
@@ -607,6 +607,18 @@ generate_js_dist_html <- function(href,
...
 
+generate_js_inline <- function(js_map) {
+  inlineJSFormatted <- paste0("<script>window.dash_clientside = window.dash_clientside || {};
+
+                              (function() {
+                              // some imaginary ES5/mergeDeep function
+                              
+                              mergeDeep(window.dash_clientside, {\n",
+                              paste0(names(js_map), ": {\n", js_map, collapse="},\n"),
+                              "});
```
alexcjohnson: Would this be easier if, like the [Python implementation](https://github.com/plotly/dash/pull/967/files#diff-228c9aefac729977642672aa1416bacaR1242-R1248), rather than saving a map you just saved individual script tags to add?
In principle it's actually nicer this way, with all of them in one script tag, but you will have to sort it out so the result is plain old js just like on the Python side.

R/dash.R
```diff
@@ -593,17 +593,47 @@ Dash <- R6::R6Class(
     # ------------------------------------------------------------------------
     callback = function(output, params, func) {
       assert_valid_callbacks(output, params, func)
-
+
```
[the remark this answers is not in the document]
rpkyle: I'm just going to write a `git commit` hook which halts on push when white space on blank lines is detected.

I'm not sure where it comes from (it's not something I've been adding), and I don't always remember to re-run my `vim` macro to cleanse these empty strings when I resave.

R/dash.R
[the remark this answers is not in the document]
rpkyle: I don't believe so. This check is to see whether one of the JavaScript callbacks has an ID (with `_dashprivate_clientside_` stripped out) that exactly matches that of an existing Dash for R callback. 

I didn't think we would want two callbacks with the same `output`, or the same set of outputs (in the multiple case) to be permitted -- one written in JavaScript and the other in R, but maybe I was wrong about that?

R/dash.R
```diff
@@ -593,17 +593,47 @@ Dash <- R6::R6Class(
...
+          private$js_map <- js_map
+        }
+        
+        func <- NULL
       } else {
         clientside_function <- func
         func <- NULL
       }
-
+
```
rpkyle: 🙄

R/dash.R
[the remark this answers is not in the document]
rpkyle: Since refactoring, this codeblock is no longer necessary and has been removed.

R/utils.R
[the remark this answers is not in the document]
rpkyle: fixed in 6704d8c

R/dash.R
```diff
@@ -772,17 +775,37 @@ Dash <- R6::R6Class(
...
+      
       if (is.function(func)) {
         clientside_function <- NULL
+      } else if (is.character(func)) {
+        # update the scripts before generating tags, and remove exact
+        # duplicates from inline_scripts
+        fn_name <- paste0("_dashprivate_", output$id)
+
+        func <- paste0('<script>\n',
+                       '                var clientside = window.dash_clientside = window.dash_clientside || {};\n',
```
HammadTheOne, line 790: Maybe @Marc-Andre-Rivet can weigh in, but should we begin to move towards `let`'s and `const`'s instead of 'var`'s for "modern" JS syntax? Although on the other hand, if we want to match the Python syntax it makes more sense to stick with `var` for now.

alexcjohnson: This code is inserted directly on the page, it's not transpiled. So unfortunately we can't safely use modern syntax here.

alexcjohnson: non-blocking, but the leading whitespace in this string is unnecessary. On the Python side, the results end up looking something like:
```js
<script>
var clientside = window.dash_clientside = window.dash_clientside || {};
var ns = clientside["_dashprivate_clientside-graph"] = clientside["_dashprivate_clientside-graph"] || {};
ns["figure"] = 
    function(data, scale) {
        return {
            'data': data,
            'layout': {
                 'yaxis': {'type': scale}
             }
        }
    }
    ;
</script>
```
But almost all the whitespace actually came from the user input. In fact if we ever decide we want to optimize bytes we can strip out all leading whitespace and carriage returns from our code and at least leading whitespace from the user-provided function string.

rpkyle: fixed in 09c73a2 and 989643a

rpkyle: I'll contribute the commit hook in a separate PR; want to test a little bit beforehand.

tests/integration/clientside/test_clientside_inline.py
```diff
@@ -0,0 +1,73 @@
...
+  function(value) {
+    sprintf("Server says %s", value)
+  }
+)
+
+app$callback(
+  output('output-clientside', 'children'),
+  params=list(input('input', 'value')),
+  "
+  function (value) {
```
Marc-Andre-Rivet, line 33: 👍

PR #142, opened 2019-11-04
Support for meta tags in Dash for R

rpkyle:
This PR proposes to add support for `<meta>` tags to Dash for R applications, as previously provided in Dash for Python via plotly/dash#286.

The newly added `generate_meta_tags` function will also check whether `http-equiv` provides the value `X-UA-Compatible` and add it for IE compatibility if it is not present. Similarly, if no character set is specified, `UTF-8` is chosen.

@jackparmer

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -1238,16 +1238,16 @@ Dash <- R6::R6Class(
       css_tags <- paste(c(css_deps,
                           css_external,
                           css_assets),
-                        collapse = "\n")
+                        collapse = "\n            ")
 
       scripts_tags <- paste(c(scripts_deps,
                               scripts_external,
                               scripts_assets,
                               scripts_invoke_renderer),
-                            collapse = "\n")
+                            collapse = "\n              ")
 
       meta_tags <- paste(generate_meta_tags(private$meta_tags),
-                         collapse = "\n")
+                         collapse = "\n            ")
       
       return(list(css_tags = css_tags,
                   scripts_tags = scripts_tags,
--- a/R/dash.R
+++ b/R/dash.R
@@ -821,6 +821,7 @@ Dash <- R6::R6Class(
     # private fields defined on initiation
     name = NULL,
     serve_locally = NULL,
+    meta_tags = NULL,
     assets_folder = NULL,
     assets_url_path = NULL,
     assets_ignore = NULL,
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -607,6 +607,28 @@ generate_js_dist_html <- function(href,
...
+generate_meta_tags <- function(metas) {
+  has_ie_compat <- any(vapply(metas, function(x) 
+    x$name == "http-equiv" && x$content == "X-UA-Compatible", 
+    logical(1)))
+  has_charset <- any(vapply(metas, function(x) 
+    "charset" %in% names(x), 
+    logical(1)))
+
+  tags <- vapply(metas, 
+                 function(tag) sprintf("<meta name=\"%s\" content=\"%s\">\n", tag$name, tag$content),
```
alexcjohnson: On the python side there are 2 differences:
- arbitrary attributes are included if they exist, rather than explicitly saying we're including name and content. This is important here for `http-equiv` (whether the specific one for IE compat or others) and `charset`, neither of which, if I'm reading correctly, will be added as expected if the default is overridden. Not sure how this would look in R, but the [Python equivalent](https://github.com/plotly/dash/blob/6c29dd0461c8cd0d8d72771a46c13cabe9851c82/dash/_utils.py#L33-L38) is:
```py
    '<meta {}>.format(
        " ".join('{}="{}"'.format(k, v) for k, v in attributes.items()),
    )
```
And if you do that, then all of the tags, including the default ones, can be formatted with the same function.
- this only matters for making a nice-looking index string, but between tags there should be not only `\n` but also as many spaces as the indentation of the first one. (in principle this gets tricky once we support index customization, you'd have to detect the indentation... but in the stock index template that looks like 12 spaces.

alexcjohnson: BTW re:
> neither of which, if I'm reading correctly, will be added as expected if the default is overridden

perhaps a couple of unit tests of the `generate_meta_tags` function?

rpkyle: > this only matters for making a nice-looking index string, but between tags there should be not only \n but also as many spaces as the indentation of the first one. (in principle this gets tricky once we support index customization, you'd have to detect the indentation... but in the stock index template that looks like 12 spaces.

Rewriting the functions to preserve line spacing will require some thought. I don't think it will be a trivial change to the code, but I might be wrong. I don't actually think we're doing this in Python either (aside from the index template), from looking at the docs index page, the tags are not auto-indented to align with the preceding lines of HTML.

I could add an argument to each of the tag generators that allows one to specify a string to collapse the tags into a single line, and insert the spacing there. I'd still have to prepend the tags with the same number of spaces, but that's simple enough. 

For custom templates, I imagine I'd need logic to extract the white space from the line containing a given interpolation key, and then I'd pass that to the tag generator.

rpkyle: > perhaps a couple of unit tests of the `generate_meta_tags` function?

integration tests added in 67fb44a

R/dash.R
```diff
@@ -1239,9 +1246,13 @@ Dash <- R6::R6Class(
                               scripts_invoke_renderer),
                             collapse = "\n")
 
+      meta_tags <- paste(generate_meta_tags(private$meta_tags),
+                         collapse = "\n")
```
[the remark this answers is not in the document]
rpkyle: fixed in d459cbe

PR #148, opened 2019-11-15
Fixes for hot reloading interval handling and refreshing apps within viewer pane

### LINKED ISSUE #146: Dash for R apps reload inconsistently in RStudio's viewer pane
Initial support for launching Dash for R apps within RStudio's viewer pane was provided in #127. Additional testing suggested that, while hot reloading worked consistently within standalone browsers, it was less reliable within the viewer pane. 

A hard reload should force a refresh on the viewer pane, just as it does within other browsers.

### LINKED ISSUE #147: Hot reloading interval control for non-asset files in app directory requires improvement
Initial hot reloading support was provided in #127, including parameters `dev_tools_hot_reload_interval` and `dev_tools_hot_reload_watch_interval`. 

The initial implementation of hot reloading did not permit a hot reload on initial update to the app itself, and performed a check based on the last event loop (cycle) end time. 

Since `fiery` defaults to updating every 0.1 seconds, this prevented hot reloads for edits to files in the application root directory, while updating assets did trigger a reload event.

A more robust implementation would track whether a hot reload has been performed at least once, and always set `permit_reload = TRUE` until one reload event has occurred. 

On subsequent cycles, it would then be possible to compare the current cycle (wall) time against the (wall) time of the previous reload event.

rpkyle:
This PR proposes several modifications to address issues noted in #146 and #147:
- a new `private$in_viewer` field tracks whether the app should be reloaded within RStudio's viewer pane
- `app_url` now uses `self$server$host` and `self$server$port`
- `self$config$hot_reload` incorrectly used `hot_reload_watch_interval`, this has been corrected to `hot_reload_interval`
- `permit_reload` is always `TRUE` until one reload event has occurred; on subsequent cycles, reloads are permitted if the current (wall) time is at least `hot_reload_watch_interval` seconds greater than the time stored in `private$last_reload`

Resolves #146, fixes #147.

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -187,6 +187,8 @@ Dash <- R6::R6Class(
       private$app_root_path <- getAppPath()
       private$app_launchtime <- as.integer(Sys.time())
       private$meta_tags <- meta_tags
+      private$has_reloaded <- FALSE
+      private$in_viewer <- FALSE
 
       # config options
       self$config$routes_pathname_prefix <- resolve_prefix(routes_pathname_prefix, "DASH_ROUTES_PATHNAME_PREFIX")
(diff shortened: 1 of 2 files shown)


R/dash.R
```diff
@@ -713,14 +715,24 @@ Dash <- R6::R6Class(
...
+          #
+          if (!is.null(private$last_cycle) & !is.null(hot_reload_watch_interval)) {
+            if (!private$has_reloaded)
+              # allow reloading on first change, then compare against the hot_reload_watch_interval
+              # to control whether future reloads are permitted on cycle end
+              permit_reload <- TRUE
+            else {
+              # determine if the time since last reload end is equal to or longer than the requested check interval
+              # (Sys.time() - private$last_reload) provides time in seconds, same units as hot_reload_watch_interval
+              permit_reload <- (Sys.time() - private$last_reload) >= hot_reload_watch_interval
```
alexcjohnson: This will work. But a common way to do it with only one variable is to start with `private$last_reload = 0` or some known distant past time (0 often works for that purpose because it typically means the unix epoch = 1/1/1970)

rpkyle: fixed in 0ce7a6d

PR #150, opened 2019-11-28
Create CODEOWNERS

Marc-Andre-Rivet:
** Follow up: Add CODEOWNERS requirement for PR to master


Marc-Andre-Rivet:
Please ignore the test failure. I'll force/merge -- DashR gets the latest version of Dash, does not support `async` and fails accordingly.

PR #157, opened 2019-12-13
Implement support for asynchronous loading in Dash for R

### LINKED ISSUE #134: Dash for R should support asynchronous component loading
As described in https://github.com/plotly/dash/pull/899, Dash for Python will soon support `eager` and `lazy` loading of components to enhance performance. Dash for R should do the same, which will require

- introducing additional logic to parse `_js_dist` and inspect the `async` field
- modifying Dash for R to store this attribute in a new (backend) field, as a logical or character string (`TRUE`, `"eager"`, `"lazy"`)
- providing support for an `eager_loading` flag
- providing support for a `_force_eager_loading` flag, e.g. https://github.com/plotly/dash-design-kit/blob/ba0b941425903fbd3d463b83480699489e8f2c91/dash_design_kit/__init__.py#L33
- implementing logic to modify component loading based on the flags above
- writing a test to validate the expected behaviour of the new implementation

Additional requirements (repasted from an earlier comment):

- Dash for R should extract `_dash_compatibility_version` from component's `__init__.py`
- may be overridden with Dash option `_dash_compatibility_version`
- warnings will be displayed to inform the user of the features inactivated by _dash_compatibility_version and which component(s) are responsible for the deprecated mode
- Need to distinguish 'lazy' files from 'dynamic' files for eagerness
- Plotly.js gets differential treatment and eagerness will not just change behavior but *also* the set of files to load

@Marc-Andre-Rivet @alexcjohnson

rpkyle:
This PR proposes to provide performance enhancements previously offered in plotly/dash#899. Briefly, it will implement the following:
- support for an `eager_loading` flag to permit automatic loading of components, as before
- selective loading of Dash components based on `eager_loading` and `async` or `dynamic` attributes for dependencies, which will optionally prevent serving of script tags in the index for certain components (thereby allowing them to load on-demand)
- fingerprinting and ETag support for caching resources
- support for `brotli`, `gzip` and `deflate` compression of text resources

Closes #134.


R/utils.R
```diff
@@ -1161,3 +1144,122 @@ dashLogger <- function(event = NULL,
...
+
+isDynamic <- function(eager_loading, resource) {
+  if (
+    is.null(resource$dynamic) && is.null(resource$async)
+  )
+    return(FALSE)
+  # need assert that async and dynamic are not both present
+  if (
+    (!is.null(resource$dynamic) && (resource$dynamic == FALSE)) ||
+    (eager_loading==TRUE && !is.null(resource$async) && (resource$async %in% c("eager", TRUE)))
```
Marc-Andre-Rivet: We could avoid the if(true) -> false and rewrite the whole func as:

```
return (!is.null(resource$dynamic) || !is.null(resource$async)) && (is.null(resource$dynamic) || resource$dynamic == TRUE) && (eager_loading == FALSE || is.null(resource$async) || !(resource$async %in% c("eager", TRUE))
```

R/utils.R
```diff
@@ -1161,3 +1144,122 @@ dashLogger <- function(event = NULL,
...
+  response$body <- paste(response$body, collapse="\n")
+  # the reqres gzip implementation requires file I/O
+  # brotli does not; when available, use brotli with
+  # a moderate level of compression for speed --
+  # the viewer pane only supports gzip and deflate,
+  # so gzip will be used when launching apps within
+  # RStudio
+  tryBrotli <- request$accepts_encoding('br')
+  if (tryBrotli == "br") {
+    response$body <- brotli::brotli_compress(charToRaw(response$body),
```
Marc-Andre-Rivet: 🥔

R/utils.R
```diff
@@ -1161,3 +1144,122 @@ dashLogger <- function(event = NULL,
...
+  extension <- getFileExt(path)
+  
+  sprintf("%s.v%sm%s.%s", 
+          file.path(dirname(path), filename),
+          gsub("[^\\w-]", "_", version, perl = TRUE),
+          hash_value,
+          extension)
+}
+
+checkFingerprint <- function(path) {
```
Marc-Andre-Rivet: Nonblocking. I was a bit confused about the fingerprint check for a few minutes because I did not see anything to handle `/` paths like in the Python impl, but it's being handled in the caller and only the filename portion of the path is passed here. Maybe renaming `path` to `filename` for clarity.

rpkyle: Actually, the path is passed here -- hence the subsequent call to `basename`, which extracts the filename from the complete path if provided. I think I'll leave `path` as-is, though the earlier call to `checkFingerprint` could be written with `path` instead of `filename` to avoid confusion.

R/utils.R
rpkyle: 🍠

PR #159, opened 2019-12-23
Added `get_asset_url()` function

### LINKED ISSUE #158: DashR does not currently have the `get_asset_url` method for loading asset URLs like Dash for Python
As originally referenced in a DashR docs issue, Dash for R currently lacks this method which is featured in the Python version. Based on the Python implementation, it would grab the asset URL from the assets directory without needing to specify the exact path. 

> @rpkyle @HammadTheOne - Do you think we could get a `get_asset_url` into Dash for R before the next release? Here is the implementation in Python:
> https://github.com/plotly/dash/blob/e622f2a31202302d0c305c56e738d1ea369403e6/dash/dash.py#L1559-L1566
> 
> If it's not in Dash for R, then we can include it in this example as a helper function.
> 

_Originally posted by @chriddyp in https://github.com/plotly/dash-enterprise-docs/pull/95_

HammadTheOne:
This PR adds in the `get_asset_url` helper function to make it easier to define and load assets within the Dash layout. It uses a combination of `regex` and mimics the functionality of the method with the same name in Dash for Python.

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -896,6 +896,35 @@ Dash <- R6::R6Class(
           message("\U{26A0} No source directory information available; hot reloading has been disabled.\nPlease ensure that you are loading your Dash for R application using source().\n")
         }
       self$server$ignite(block = block, showcase = showcase, ...)
+      },
+    
+    # ------------------------------------------------------------------------
+    # get_asset_url helper function
+    # ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


HammadTheOne:
As per #158, different implementation using unexported function in `utils.R` will be used instead, but using the names attribute of the asset map.

PR #160, opened 2019-12-27
Support returning asset URLs via public method within Dash class

rpkyle:
The previous implementation of `get_asset_url` in Dash for R was not particularly useful -- it required passing an object whose asset paths were set on the `names` attribute of the object, which is used internally by Dash, but not generally available to application developers. It was also unexported, and not used elsewhere in the package -- effectively dead code.

This revision makes the `get_asset_url` a public method of the Dash class itself, so it is available to developers for use. 

If this PR is merged, the method will permit returning the URL of any file in the assets directory, provided that a valid relative path within an app's `assets_folder` is passed. If the asset is not found, an error is returned.


R/dash.R
```diff
@@ -702,7 +702,43 @@ Dash <- R6::R6Class(
...
+    # ------------------------------------------------------------------------
+    # return asset URLs
+    # ------------------------------------------------------------------------
+    get_asset_url = function(asset_path, prefix = "/") {
+      asset <- lapply(private$asset_map, 
+                      function(x) {
+                        # asset_path should be prepended with the full app root & assets path
+                        # if leading slash(es) present in asset_path, remove them before
+                        # assembling full asset path
+                        asset_path <- file.path(dirname(private$app_root_path),
```
HammadTheOne: I encountered the following error when running this line to generate the `asset_path`:

`Error in dirname(private$app_root_path) : a character vector argument expected`

Digging a bit deeper into it, the `app_root_path` private field in Dash calls  the function `getAppPath` from `utils.R` upon initialization of the Dash class, and returns the value `FALSE` if the app isn't run via `source()`. 

A potential fix could be:
`ifelse(is.logical(private$app_root_path), "", dirname(private$app_root_path))`, but I'm not exactly sure what needs to be there when the condition is TRUE to get the correct path. 

A more permanent fix would be to tweak the `getAppPath` function to deliver a more comprehensive return value for the `else` condition.

rpkyle:
> Hey Ryan, I didn't see any real issues with the code, it's written concisely and makes sense, the comments make it easy to follow.
> 
> The `get_asset_url` method works perfectly fine when running the app from `source('app.R')`, but has an issue when run from the script. Once that's figured out, it should be good to go.

Hmm, yes. This is going to be tricky to resolve. We don't have a reliable way to determine the application root unless code is executed using `source()`. Basically, `source()` calls `parse()`, which populates `srcref` attributes, as described here:

https://journal.r-project.org/archive/2010-2/RJournal_2010-2_Murdoch.pdf

The `srcref` attributes are very handy, since they are a relatively safe way of identifying the location of R code. 

Otherwise, we're basically making our best guess at the value of `app_root_path` whenever code is pasted/entered into the console.

@alexcjohnson Any suggestions? We're trying to reproduce the behaviour of `get_asset_url` in R to mimic that in Python, but running into a snag when code is not loaded via `source()`.

alexcjohnson:
> Otherwise, we're basically making our best guess at the value of `app_root_path` whenever code is pasted/entered into the console.

This concern isn't new in this PR, right? So we can make an issue for it but I wouldn't try to address it in this PR.

When you paste code in the console there's really no reason to expect it comes from *anywhere* in the filesystem. So it seems to me in fact a bad idea to try to impute some location to it. If users want to work this way, the root path would need to be provided explicitly in the code; otherwise we should steer folks to a workflow in which the app is invoked explicitly via its path.

PR #165, opened 2020-01-13
Autoset routes and requests pathname prefixes if DASH_APP_NAME is set

rpkyle:
This PR proposes to modify the resolution logic for `resolve_prefix`, which is called by Dash when setting `self$routes_pathname_prefix` and `self$requests_pathname_prefix`:
- check if `routes_pathname_prefix` or `requests_pathname_prefix` has been passed to Dash for R directly, and use if available
- check if `DASH_ROUTES_PATHNAME_PREFIX` or `DASH_REQUESTS_PATHNAME_PREFIX` have been set in current environment, and use if available
- check if `DASH_APP_NAME` has been set in environment, and use if available
- check if `DASH_URL_BASE_PATHNAME` has been set in environment, and use if available
- return the value of `base_pathname` if resolution fails as directed above

In addition, the function name is camelCased instead of using snake case.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -468,7 +468,7 @@ resolvePrefix <- function(prefix, environment_var, base_pathname) {
     env_base_pathname <- Sys.getenv("DASH_URL_BASE_PATHNAME")
     app_name <- Sys.getenv("DASH_APP_NAME")
     
-    else if (prefix_env != "")
+    if (prefix_env != "")
       return(prefix_env)
     else if (app_name != "")
       return(sprintf("/%s/", app_name))
--- a/R/utils.R
+++ b/R/utils.R
@@ -457,7 +457,7 @@ valid_seq <- function(params) {
   }
 }
 
-resolve_prefix <- function(prefix, environment_var, base_pathname) {
+resolvePrefix <- function(prefix, environment_var, base_pathname) {
   if (!(is.null(prefix))) {
     assertthat::assert_that(is.character(prefix))
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -463,16 +463,19 @@ resolve_prefix <- function(prefix, environment_var, base_pathname) {
 
     return(prefix)
   } else {
+    # Check environment variables
     prefix_env <- Sys.getenv(environment_var)
-    if (prefix_env != "") {
+    env_base_pathname <- Sys.getenv("DASH_URL_BASE_PATHNAME")
+    app_name <- Sys.getenv("DASH_APP_NAME")
+    
+    else if (prefix_env != "")
       return(prefix_env)
-    } else {
-      env_base_pathname <- Sys.getenv("DASH_URL_BASE_PATHNAME")
-      if (env_base_pathname != "")
-        return(env_base_pathname)
-      else
-        return(base_pathname)
-    }
+    else if (app_name != "")
+      return(sprintf("/%s/", app_name))
+    else if (env_base_pathname != "")
+      return(env_base_pathname)
+    else
+      return(base_pathname)
   }
 }
 
--- a/R/dash.R
+++ b/R/dash.R
@@ -213,8 +213,8 @@ Dash <- R6::R6Class(
       private$in_viewer <- FALSE
 
       # config options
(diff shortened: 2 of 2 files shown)


R/utils.R
```diff
@@ -457,22 +457,25 @@ valid_seq <- function(params) {
...
+    else if (prefix_env != "")
       return(prefix_env)
-    } else {
-      env_base_pathname <- Sys.getenv("DASH_URL_BASE_PATHNAME")
-      if (env_base_pathname != "")
-        return(env_base_pathname)
-      else
-        return(base_pathname)
-    }
+    else if (app_name != "")
```
HammadTheOne: Seems to be a typo here with the opening braces for the `else` statement that's preventing the package from installing.

R/utils.R
```diff
@@ -457,22 +457,25 @@ valid_seq <- function(params) {
...
 
     return(prefix)
   } else {
+    # Check environment variables
     prefix_env <- Sys.getenv(environment_var)
-    if (prefix_env != "") {
+    env_base_pathname <- Sys.getenv("DASH_URL_BASE_PATHNAME")
+    app_name <- Sys.getenv("DASH_APP_NAME")
+    
+    else if (prefix_env != "")
```
HammadTheOne:
```suggestion
    if (prefix_env != "")
```

R/utils.R
```diff
@@ -457,22 +457,25 @@ valid_seq <- function(params) {
...
 
     return(prefix)
   } else {
+    # Check environment variables
     prefix_env <- Sys.getenv(environment_var)
-    if (prefix_env != "") {
+    env_base_pathname <- Sys.getenv("DASH_URL_BASE_PATHNAME")
+    app_name <- Sys.getenv("DASH_APP_NAME")
+    
+    if (prefix_env != "")
```
HammadTheOne: I made a slight change here to fix the conditional statement.

rpkyle:
> There was a slight issue with a conditional but that's been sorted out with the last commit. I did notice however, that the changes aren't backwards compatible, I had to remove the references to some of the environment variables declared in previous versions of the deployment template, but I think (based on our discussion earlier) that that's by design and should be fine going forward with future apps/deployments.

Thanks @HammadTheOne. There are definitely some changes introduced as a result of all this work, but the end effect is a smoother, simpler deployment process overall.

R/utils.R
rpkyle: 🤦‍♂ Yep, one too many `else if` statements here.

chriddyp:
> I did notice however, that the changes aren't backwards compatible

I find this surprising. Which line of code is backwards incompatible? I would've expected that if these variables are _already_ set (as in our current examples), they would continue to be used.

HammadTheOne:
I believe the line looking for the `appName` was the one, here's the logs that popped up before I removed those lines:

```
        R version 3.6.2 (2019-12-12) -- "Dark and Stormy Night"
        Copyright (C) 2019 The R Foundation for Statistical Computing
        Platform: x86_64-pc-linux-gnu (64-bit)
        R is free software and comes with ABSOLUTELY NO WARRANTY.
        You are welcome to redistribute it under certain conditions.
        Type 'license()' or 'licence()' for distribution details.
        R is a collaborative project with many contributors.
        Type 'contributors()' for more information and
        'citation()' on how to cite R or R packages in publications.
        Type 'demo()' for some demos, 'help()' for on-line help, or
        'help.start()' for an HTML browser interface to help.
        Type 'q()' to quit R.
        > appName <- Sys.getenv("DASH_APP_NAME")
        > 
        > if (appName != "") {
        +   pathPrefix <- sprintf("/%s/", appName)
        +   
        +   Sys.setenv(DASH_ROUTES_PATHNAME_PREFIX = pathPrefix,
        +              DASH_REQUESTS_PATHNAME_PREFIX = pathPrefix,
        +              )
        + }
        Error in Sys.setenv(DASH_ROUTES_PATHNAME_PREFIX = pathPrefix, DASH_REQUESTS_PATHNAME_PREFIX = pathPrefix,  : 
          argument is missing, with no default
        Execution halted
 =====> end ctest web container output
```

PR #167, opened 2020-01-13
Remove DASH_HOST and DASH_PORT, use std vars

rpkyle:
This PR proposes to replace `DASH_HOST` and `DASH_PORT` with `HOST` and `PORT`, since the former were never actively used in practice, and for parity with Dash for Python.


josegonzalez:
Is it okay to have this BC break?

alexcjohnson:
Yes, now's the time for breaking compatibility in the name of standardization.

PR #168, opened 2020-01-14
Support for index page templating

### LINKED ISSUE #42: More flexible index page templating
Issue: https://github.com/plotly/dash/issues/265
PR: https://github.com/plotly/dash/pull/286

HammadTheOne:
This PR adds in custom index functionality by adding in a method to the Dash class which takes in an HTML index string with some necessary defaults, and whatever custom additions or changes a user wants to make, and then renders it above or below the app based on where the code is added. Custom `meta_tags`, `title`, and scripts like the Google Tag Manager analytics can be added to the index body and loaded with the `script` tag. 

This functionality brings Dash for R to parity with the major feature set of Dash for Python.

An example of a custom string is below:

```
string <-
"<!DOCTYPE html>
        <html>
          <head>
            {meta_tags}
            <title>{private$name}</title>
            {favicon}
            {css_tags}
            <script type ='text/javascript' src='googleanalytics.js'></script>
          </head>
          <body>
            <p>'This is a custom string!'</p>
            <!-- Google Tag Manager (noscript) -->
              <noscript><iframe src='https://www.googletagmanager.com/ns.html?id=GTM-XXXXXXX'
              height='0' width='0' style='display:none;visibility:hidden'></iframe></noscript>
          <!-- End Google Tag Manager (noscript) -->
            <div id='react-entry-point'>
              <div class='_dash-loading'>Loading...</div>
            </div>
            <footer>
              <script id='_dash-config' type='application/json'> {to_JSON(self$config)} </script>
              {scripts_tags}
            </footer>
          </body>
        </html>"
```

Closes #42


R/dash.R
```diff
@@ -505,55 +701,280 @@ Dash <- R6::R6Class(
...
+    
+    # ------------------------------------------------------------------------
+    # specify a custom index string
+    # ------------------------------------------------------------------------
+    index_string = function(string) {
+      if (!grepl("react-entry-point", string)) stop("React entry point div must be defined within the index string.")
+      if (!grepl("_dash-config", string)) stop("Dash config script has not been defined within the index string.")
+      if (!grepl("scripts_tags", string)) stop("{scripts_tags} must be included to render the Dash app and any JavaScript files in /assets/.")
+      private$custom_index = string
+    },
```
rpkyle: On the Python side, it looks like the analogous code block is

```
    @index_string.setter
    def index_string(self, value):
        checks = (
            (_re_index_entry.search(value), "app_entry"),
            (_re_index_config.search(value), "config"),
            (_re_index_scripts.search(value), "scripts"),
        )
        missing = [missing for check, missing in checks if not check]
        if missing:
            raise exceptions.InvalidIndexException(
                "Did you forget to include {} in your index string ?".format(
                    ", ".join("{%" + x + "%}" for x in missing)
                )
            )
        self._index_string = value
```

We can leave the (internal) variable names as they are for now, but the interface should probably match the Python side, so that we're looking for `app_entry`, `config`, and `scripts`. 

It would also be good if we could throw an error message which includes all the elements that are missing, as Python also does -- that way if 2/3 or 3/3 need fixing, the user will be notified of all of them. (Since here if both `_dash-config` and `react-entry-point` are not defined, you'll only know that the latter is wrong, then get prompted again that `_dash-config` must be defined too.)

Here's an existing example that does something similar (in terms of returning an message with a variable number of errors):

https://github.com/plotly/dashR/blob/f81fcb5332eb66a0933d3fd5b1a404279fb3c459/R/utils.R#L290-L300

R/dash.R
```diff
@@ -1439,8 +1461,14 @@ Dash <- R6::R6Class(
       # insert meta tags if present
       meta_tags <- all_tags[["meta_tags"]]
 
-      private$.index <- sprintf(
-        '<!DOCTYPE html>
+      if (!is.null(private$custom_index)) {
+        interpolated_index <- glue::glue(private$custom_index)
```
rpkyle: @HammadTheOne To be honest, I'm not quite sure what this call to `glue::glue` is doing. 

If it's simple concatenation, I'd replace with a call to `paste0` or `sprintf` to avoid pulling in the dependency. Something like

```
> test <- c("foo", "bar")
paste0("A few things to worry about: ", paste0(test, collapse=", "))
[1] "A few things to worry about: foo, bar"
```

or

```
sprintf("A few things to worry about: %s", paste(test, collapse = ", "))
```

...though generally we try to avoid mixing `sprintf` and `paste`.

R/dash.R
```diff
@@ -1439,8 +1461,14 @@ Dash <- R6::R6Class(
...
+        
+        private$.index <- interpolated_index
```
rpkyle: One thing I don't think I see here that we likely want to implement are the interpolation keys -- see the comments in this merged PR for more info:

https://github.com/plotly/dash/pull/286

Let me know if I can help with anything here, but the general goal is to make the implementation as similar as we can to what currently exists on the Python side (within reason!).

R/dash.R
HammadTheOne: In this case, I was thinking of using `glue` since it can handle string interpolation for us, and is syntactically similar to the way `index_string` is handled in the Python implementation using braces for the required keys and adding in the `meta_tags`, `scripts_tags` variables that are already declared for the default index. 

Based on that, is it still something that would be better changed to base R functions or should it remain?

.circleci/config.yml
```diff
@@ -7,6 +7,9 @@ jobs:
     working_directory: ~/dashr
     docker:
       - image: byronz/dashr:ci
+        environment:
```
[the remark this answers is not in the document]
HammadTheOne: That makes sense, and I think we also wanted to move away from personal hosted builds, so I think we can sync that up.

DESCRIPTION
```diff
@@ -1,19 +1,19 @@
 Package: dash
 Title: An Interface to the Dash Ecosystem for Authoring Reactive Web Applications
-Version: 0.1.0
+Version: 0.2.0
```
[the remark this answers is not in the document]
HammadTheOne: Fixed in 3aa0f5.

DESCRIPTION
```diff
@@ -1,19 +1,19 @@
...
 Authors@R: c(person("Chris", "Parmer", role = c("aut"), email = "chris@plot.ly"), person("Ryan Patrick", "Kyle", role = c("aut", "cre"), comment = c(ORCID = "0000-0001-5829-9867"), email = "ryan@plot.ly"), person("Carson", "Sievert", role = c("aut"), comment = c(ORCID = "0000-0002-4958-2844")), person(family = "Plotly Technologies", role = "cph")) 
 Description: A framework for building analytical web applications, Dash offers a pleasant and productive development experience. No JavaScript required.
 Depends:
   R (>= 3.0.2)
 Imports:
-  dashHtmlComponents (== 1.0.0),
-  dashCoreComponents (== 1.0.0),
-  dashTable (== 4.0.2),
+  dashHtmlComponents (== 1.0.2),
```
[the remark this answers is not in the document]
HammadTheOne: Noted - Will do when other other comments are resolved and we have concrete versions for 0.3.0.

DESCRIPTION
```diff
@@ -31,13 +31,13 @@ Collate:
   'imports.R'
   'print.R'
   'internal.R'
-Remotes: plotly/dash-html-components@17da1f4,
-  plotly/dash-core-components@cc1e654,
-  plotly/dash-table@042ad65,
-  plotly/reqres@55072ab
+Remotes: plotly/dash-html-components@55c3884,
```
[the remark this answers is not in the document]
HammadTheOne: Same as above, I think when we're about to merge for a release we can update these to the most recent commit hashes.

NAMESPACE
```diff
@@ -26,4 +27,4 @@ importFrom(routr,RouteStack)
 importFrom(routr,ressource_route)
 importFrom(stats,setNames)
 importFrom(tools,file_ext)
-importFrom(utils,getFromNamespace)
+importFrom(utils,getFromNamespace)
```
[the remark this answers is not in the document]
HammadTheOne: Fixed in a2410d.

R/dash.R
```diff
@@ -76,26 +87,74 @@
...
+#'   }
+#'   \item{`run_server(host =  Sys.getenv('DASH_HOST', "127.0.0.1"),
 #'    port = Sys.getenv('DASH_PORT', 8050), block = TRUE, showcase = FALSE, ...)`}{
-#'     Launch the application. If provided, `host`/`port` set
-#'     the `host`/`port` fields of the underlying [fiery::Fire] web
-#'     server. The `block`/`showcase`/`...` arguments are passed along
+#'     The `run_server` method has 13 formal arguments, several of which are optional:
+#'     \describe{
+#'       \item{host}{Character. A string specifying a valid IPv4 address for the Fiery server, or `0.0.0.0` to listen on all addresses. Default is `127.0.0.1` Environment variable: `DASH_HOST`.}
+#'       \item{port}{Integer. Specifies the port number on which the server should listen (default is `8050`). Environment variable: `DASH_PORT`.}
```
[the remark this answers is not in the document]
HammadTheOne: Should be fixed with a158a3.

R/dash.R
```diff
@@ -76,26 +87,74 @@
...
+#'   \item{`get_asset_url(asset_path, prefix)`}{
+#'     The `get_asset_url` method permits retrieval of an asset's URL given its filename.
+#'     For example, `app$get_asset_url('style.css')` should return `/assets/style.css` when
+#'     `assets_folder = 'assets'`. By default, the prefix is the value of `requests_pathname_prefix`,
+#'     but this is configurable via the `prefix` parameter. Note: this method will
+#'     present a warning and return `NULL` if the Dash app was not loaded via `source()`
+#'     if the `DASH_APP_PATH` environment variable is undefined.
+#'   }
+#'  \item{`index_string(string)`}{
+#'     The `index_string` method allows the specification of a custom index by changing
```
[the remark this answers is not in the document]
HammadTheOne: Fixed in 17a61d, the entire key has been moved to just a simple {config} key, so I'll remove that from the description as well (13ebec).

R/dash.R
```diff
@@ -147,16 +209,22 @@ Dash <- R6::R6Class(
...
+      private$compress <- compress
+      private$app_root_path <- getAppPath()
+      private$app_launchtime <- as.integer(Sys.time())
+      private$meta_tags <- meta_tags
+      private$in_viewer <- FALSE
 
       # config options
-      self$config$routes_pathname_prefix <- resolve_prefix(routes_pathname_prefix, "DASH_ROUTES_PATHNAME_PREFIX")
-      self$config$requests_pathname_prefix <- resolve_prefix(requests_pathname_prefix, "DASH_REQUESTS_PATHNAME_PREFIX")
+      self$config$routes_pathname_prefix <- resolve_prefix(routes_pathname_prefix, "DASH_ROUTES_PATHNAME_PREFIX", url_base_pathname)
```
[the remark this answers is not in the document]
HammadTheOne: Fixed with 098466.

R/dash.R
HammadTheOne: Addressed in 1138e4. Might be worth taking another look at again though, just to make sure the index key changes are the ones we want to stick with moving forward, this did involve changes to the default index as well.

Also made another change in 17a61d, this moves the `react-entry-point` div to the {app_entry} key itself. From a user standpoint it's better because it makes the custom index less complex and easier to manipulate without breaking the essential components.

R/dash.R
HammadTheOne: Implemented in 036e4e. It's not as clean as I'd want it to be, but I'd be interested to hear your thoughts on it.

R/dash.R
rpkyle: @HammadTheOne I think this is OK here, especially given that we're already pulling in this package because of another Dash dependency.

rpkyle: @HammadTheOne Actually, upon reflection, I don't think we should use `glue::glue` as-is, because of something noted here previously:

https://github.com/plotly/dash/pull/286#discussion_r201783101

We want the keys to be enclosed with `%`, and I think `glue` can't handle that, since it parses their contents as executable R code. Typically `%` in R is used to enclose operators to be applied to matrices, e.g. `%*%` for matrix multiplication. So `glue` will fail:

```
> mykey <- "test"
> glue::glue("this is a {mykey}")
this is a test
> glue::glue("this is a {%mykey%}")
Error in parse(text = text, keep.source = FALSE) : 
  <text>:1:1: unexpected SPECIAL
1: %mykey%
    ^
```

HammadTheOne: I think this is a great change for parity, that way people can even use the same index string between Python and R apps if they want. Added/fixed in 4023b6.

R/dash.R
```diff
@@ -763,11 +811,53 @@ Dash <- R6::R6Class(
...
+    interpolate_index = function(template_index = private$template_index[[1]], ...) {
+      template = template_index
+      kwargs <- list(...)
+      
+      for (name in names(kwargs)) {
+        key = paste0('\\{\\%', name, '\\%\\}')
+        template = sub(key, kwargs[[name]], template)
+      } 
+      
+      cat(template)
```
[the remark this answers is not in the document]
rpkyle: Fixed in a0c2f7e

R/dash.R
```diff
@@ -763,11 +811,53 @@ Dash <- R6::R6Class(
...
+      
+      requiredKeys <- c("app_entry", "config", "scripts")
+      
+      checks <- sapply(requiredKeys, function(x) grepl(x, names(kwargs)))
```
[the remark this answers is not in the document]
rpkyle: I think I found a way to make this a little more concise in 5ae7549.

R/dash.R
rpkyle: I've tried to make the implementation a little more concise while aiming for 🌴 code, in 5ae7549 and 39d52e0.

DESCRIPTION
```diff
@@ -32,8 +32,8 @@ Collate:
   'print.R'
   'internal.R'
 Remotes: plotly/dash-html-components@55c3884,
-  plotly/dash-core-components@c107e0f,
-  plotly/dash-table@3058bd5
+  plotly/dash-core-components@7fa7feb,
```
Marc-Andre-Rivet: I think this needs to point here instead
https://github.com/plotly/dash-core-components/commit/fc153b4e8b980231be3d49c0f20281b2fc9678cf

PR #176, opened 2020-02-12
Add unit tests for index customization

rpkyle:
This PR proposes to introduce some basic unit tests for the index customization feature [#168]  recently added to Dash for R.

@HammadTheOne


tests/testthat/test-index.R
```diff
@@ -0,0 +1,51 @@
...
+  <body>
+  {%app_entry%}
+  <footer>
+  {%config%}
+  {%scripts%}
+  </footer>
+  </body>
+  </html>"
+  
+  app <- Dash$new()
```
Marc-Andre-Rivet: I'm assuming meta_tags, favicon, css_tag are optional? Might as well test that too for completeness.

rpkyle: Indeed, they are optional, but can test them easily enough, will add.

rpkyle: Actually, those go in the integration test, since they're tested with Selenium. Out of scope for this PR, but we can consider them for the `pytest` script.

PR #180, opened 2020-02-19
Replace all instances of plot.ly with plotly.com.

shammamah:
### About 
Related to https://github.com/plotly/dash-core/issues/152

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -668,7 +668,7 @@ encode_plotly <- function(layout_objs) {
         "x" %in% names(layout_objs) &&
         any(c("visdat", "data") %in% names(layout_objs$x))) {
       # check to determine whether the current element is an
-      # object output from the plotly.com or ggplotly function;
+      # object output from the plot_ly or ggplotly function;
       # if it is, we can safely assume that it contains no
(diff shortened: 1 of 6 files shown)


R/utils.R
```diff
@@ -668,9 +668,9 @@ encode_plotly <- function(layout_objs) {
         "x" %in% names(layout_objs) &&
         any(c("visdat", "data") %in% names(layout_objs$x))) {
       # check to determine whether the current element is an
-      # object output from the plot_ly or ggplotly function;
+      # object output from the plotly.com or ggplotly function;
```
rpkyle:
```suggestion
      # object output from the plot_ly or ggplotly function;
```

R/utils.R
```diff
@@ -668,9 +668,9 @@ encode_plotly <- function(layout_objs) {
...
       # if it is, we can safely assume that it contains no
-      # other plot_ly or ggplotly objects and return the updated
+      # other plotly.com or ggplotly objects and return the updated
```
rpkyle:
```suggestion
      # other plot_ly or ggplotly objects and return the updated
```

Issue #185, opened 2020-03-16
Ensure that Dash for R properly supports arbitary file extensions within components


rpkyle:
As provided by plotly/dash#481 for the Python version of the backend, Dash for R should support a range of mimetypes for component library assets, such as fonts (for example).

rpkyle:
Resolved by #186; closing.

PR #186, opened 2020-03-17
Provide support for arbitrary file extensions within Dash for R component libraries

rpkyle:
This PR proposes to address #185, which implements functionality found in plotly/dash#481. In particular, the internal `get_mimetype` function should now handle both `.map` and `.js.map`, and uses `guess_mimetype` from the `mime` package for extensions other than `.js`, `.css` and `.map`/`.js.map`.

This feature will require edits to Dash's package generator; a PR proposing those changes is in process.

If the extension is unknown, Dash for R will assume `application/octet-stream` instead. Other features this PR would provide:
- support for loading binary file dependencies within the `_dash-component-suites` handler
- support for properly setting a variety of new `resource$type` values via the `mime` package

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
           command: |
             python -m venv venv
             . venv/bin/activate
-            git clone --depth 1 https://github.com/plotly/dash.git dash-main
+            git clone -b 481-arbitrary-extensions --depth 1 https://github.com/plotly/dash.git dash-main
             cd dash-main && pip install -e .[dev,testing] --progress-bar off && cd ..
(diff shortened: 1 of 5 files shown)


rpkyle:
This PR is currently on hold until minor edits are made within Dash for component library assets, as noted in https://github.com/plotly/dash/pull/1186.

.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           command: |
             python -m venv venv
             . venv/bin/activate
-            git clone --depth 1 https://github.com/plotly/dash.git dash-main
+            git clone -b 481-arbitrary-extensions --depth 1 https://github.com/plotly/dash.git dash-main
```
Marc-Andre-Rivet, line 40: Temporary, for (1) the modified standard test component, (2) the modified generator

PR #194, opened 2020-04-27
Add show_undo_redo parameter to Dash for R

rpkyle:
This pull request proposes to enable support for displaying the "undo" and "redo" button to step through the history of the app state for parity with Dash for Python. A simple integration test is also added to test functionality.

The "undo" button will appear like this:
Screenshot 2020-04-27 10 11 18

... and the "redo" button will appear like this:
Screenshot 2020-04-27 10 11 32


R/dash.R
```diff
@@ -305,6 +311,7 @@ Dash <- R6::R6Class(
       self$config$requests_pathname_prefix <- resolvePrefix(requests_pathname_prefix, "DASH_REQUESTS_PATHNAME_PREFIX", url_base_pathname)
       self$config$external_scripts <- external_scripts
       self$config$external_stylesheets <- external_stylesheets
+      self$config$show_undo_redo <- show_undo_redo
```
alexcjohnson, line 314: Ah ok - you're calling `to_JSON(self$config)` for the `_dash-config` section of the index, so putting this into `self$config` is enough to get it included in the index. That probably means there's too much stuff being shoved in there. Doesn't have to be addressed now, feel free to make a new issue for it, but on the Py side we explicitly set what makes its way to the front end [here](https://github.com/plotly/dash/blob/680423e4962f508d5f9101d5752b30b81988b0b6/dash/dash.py#L490-L509)

PR #197, opened 2020-05-08
Fix bug which prevented multiple outputs in callbacks returning Dash components in R

rpkyle:
This PR proposes a minor modification to the way Dash for R's `_dash-update-component` handler resolves multiple outputs in callbacks which return one or more Dash components. This bug was discovered accidentally while fixing another bug in the Dash for R docs.

Previously, `updatable_outputs` implicitly assumed `output_value`'s elements were members of only one class, since the specified return type of `vapply` was `character(1)`:

https://github.com/plotly/dashR/blob/b3be88ddf5cc3b3b0cc5edd063068779638680ab/R/dash.R#L477-L478

This was insufficiently robust, since valid return values from Dash callbacks can be objects in R which are members of more than one class.

As an example, if a callback returns `list(htmlDiv(), "")`, the class membership of its elements would be 

```
> lapply(list(htmlDiv(), ""), class)
[[1]]
[1] "dash_component" "list"          

[[2]]
[1] "character"
```

The original definition of `updatable_outputs` would essentially attempt `vapply(list(htmlDiv(), ""), class, character(1))` which produces the cryptic error

```
Error in vapply(list(htmlDiv(), ""), class, character(1)) : 
  values must be length 1,
 but FUN(X[[1]]) result is length 2
```

By rewriting the original line as

```
updatable_outputs <- vapply(output_value, function(x) !("no_update" %in% class(x)), logical(1))
```

class membership is inspected more flexibly, and a logical result is returned in the `vapply` step instead of a `character` vector of class names.

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -399,7 +399,6 @@ Dash <- R6::R6Class(
       dash_update <- paste0(self$config$routes_pathname_prefix, "_dash-update-component")
       route$add_handler("post", dash_update, function(request, response, keys, ...) {
         request <- request_parse_json(request)
-
         if (!"output" %in% names(request$body)) {
           response$body <- "Couldn't find output component in request body"
           response$status <- 500L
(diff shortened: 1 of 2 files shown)


rpkyle:
@alexcjohnson test for callbacks returning components added in eea0b46.

PR #198, opened 2020-05-08
Permit returning components from multiple output callbacks 

rpkyle:
This PR merges #197 into `master` to resolve a bug with the way multiple outputs were handled in callbacks returning components.


rpkyle:
Percy temporarily disabled; CircleCI passing now. Will re-enable.

PR #200, opened 2020-05-15
Update title to be Dash instead of dash

chriddyp:
Convention in browser tabs is for the title to be upper case, `dash` is the odd one out right now

image


rpkyle:
@chriddyp Looks good, I'll fix the broken test now so that we can merge this one.

PR #202, opened 2020-05-25
Remove timestamps from asset URLs

### LINKED ISSUE #201: Remove outdated timestamp for Dash assets
Dash for R still retains `?v=` and `&m=` within `utils.R`, when serving fingerprinted assets via the `_dash-component-suites` handler:
https://github.com/plotly/dashR/blob/9c5a30760e3be985c866f122e9c2bd6ce2a516a6/R/utils.R#L167-L182

This has since been eliminated in Dash for Python. For the sake of parity, it should be excised from Dash for R also.

@Marc-Andre-Rivet

rpkyle:
Closes #201 .

TODO:
- [x] update regex for parsing URLs, per https://github.com/plotly/dashR/issues/201#issuecomment-631075481


rpkyle:
@Marc-Andre-Rivet All references to `?m=` have been expunged from Dash for R, yet this error appears on soft reloading CSS:

I believe the cause is L35745 within `dash_renderer.dev.js`, which appends `?m=` with the CSS modification timestamp to the request coming in to the backend from the renderer:

```javascript
if (a.modified > 0) {
  var link = document.createElement('link');
  link.href = "".concat(a.url, "?m=").concat(a.modified);
  link.type = 'text/css';
  link.rel = 'stylesheet';

  this._head.appendChild(link); // Else the file was deleted.
}
```

Is this also related to [`flask.Blueprint`](https://github.com/plotly/dash/blob/ebac7d4e82ebb7895b3437443863a889f1d30587/dash/dash.py#L385)? Or can we eliminate the `concat` statements now?

Until 👆 is resolved and the soft reloading test passes again, this PR will have to be held.

rpkyle:
@Marc-Andre-Rivet I've opened an issue to track the above, since I hadn't previously: https://github.com/plotly/dash/issues/1359. I can take this one on if it would be helpful.

rpkyle:
It turns out that the underlying code which takes advantage of `?m` is a bit more complex to remove based on the bug observed in https://github.com/plotly/dash/issues/1363; this will likely be held until some housekeeping is done in Dash for Python as well.

PR #205, opened 2020-05-28
Dash for R v0.5.0

rpkyle:
### Added
- Dash for R now depends on the `brotli` package explicitly; previously it was loaded when importing `reqres`. [#204](https://github.com/plotly/dashR/pull/204)

### Changed
- Dash for R no longer wraps the layout in an `htmlDiv` internally, for parity with Dash for Python. Starting in v0.5.0, the `layout` method only accepts a single argument, and that argument must be a Dash component or a function that returns a Dash component. [#121](https://github.com/plotly/dashR/pull/121)
- Package documentation has been significantly refactored to use new features of `roxygen2` when documenting R6 classes
- The `title` method now specifies `Dash` as the default application title instead of `dash`. [#200](https://github.com/plotly/dashR/pull/200)

### Fixed
- A minor bug in `validate_keys` which prevented `interpolate_index` from working as intended has been resolved


CHANGELOG.md
```diff
@@ -1,6 +1,18 @@
 # Change Log for Dash for R
 All notable changes to this project will be documented in this file.
 
+## [0.5.0 ] - 2020-05-28
+### Added
+- Dash for R now depends on the `brotli` package explicitly; previously it was loaded when importing `reqres`. [#204](https://github.com/plotly/dashR/pull/204)
+
+### Changed
+- Dash for R no longer wraps the layout in an `htmlDiv` internally, for parity with Dash for Python. Starting in v0.5.0, the `layout` method only accepts a single argument, and that argument must be a Dash component or a function that returns a Dash component. [#121](https://github.com/plotly/dashR/pull/121)
```
josegonzalez, line 9: What does this mean for the majority of apps? Any changes folks will need to do to make sure their apps don't do something odd?

.circleci/config.yml
```diff
@@ -24,7 +24,6 @@ jobs:
             echo "RUNNING JOB: ${CIRCLE_JOB}"
             echo "JOB PARALLELISM: ${CIRCLE_NODE_TOTAL}"
             echo "CIRCLE_REPOSITORY_URL: ${CIRCLE_REPOSITORY_URL}"
-            export PERCY_ENABLE=0
```
josegonzalez, line 27: Might want to add this as an enhancement :D

R/utils.R
```diff
@@ -338,8 +329,8 @@ insertIntoCallbackMap <- function(map, inputs, output, state, func, clientside_f
                                           clientside_function=clientside_function
                                           )
   if (length(map) >= 2) {
-    ids <- lapply(names(map), function(x) dash:::getIdProps(x)$ids)
-    props <- lapply(names(map), function(x) dash:::getIdProps(x)$props)
+    ids <- lapply(names(map), function(x) getIdProps(x)$ids)
+    props <- lapply(names(map), function(x) getIdProps(x)$props)
```
josegonzalez, line 333: Why the namespace removal?

R/utils.R
```diff
@@ -1341,10 +1332,13 @@ interpolate_str <- function(index_template, ...) {
   return(template)
 }
 
-validate_keys <- function(string) {
+validate_keys <- function(string, is_template) {
```
josegonzalez, line 1335: Does the signature change impact existing calls to `validate_keys` that only have a single argument?

tests/testthat/test-index.R
```diff
@@ -41,12 +41,3 @@ test_that("Omitting required template keys produces warnings", {
     "Did you forget to include config in your index string?"
   )
 })
-
-test_that("Customizing title using `name` produces a warning", {
```
josegonzalez, line 45: Should we raise an error now instead?

CHANGELOG.md
rpkyle: @josegonzalez Good question; ultimately, it means that code like this:

```r
app$layout(
  htmlDiv("Some text"),
  htmlDiv("Some more text")
)
```

is no longer treated as valid. It now needs to be wrapped within a div, with the children passed as a list:

```r
app$layout(
  htmlDiv(
    list(
      htmlDiv("Some text"),
      htmlDiv("Some more text")
    )
  )
)
```

This was an artifact of the initial implementation (from the very, very beginning of Dash for R), which supported passing components as `...` (like the R equivalent of `kwargs`). Realistically it's not supported by Dash, but was _possible_ in Dash for R previously.

I'd consider this a breaking change-level modification, but now is the time to :hocho: the offending code, before we reach v1.0.

.circleci/config.yml
rpkyle: 😹 I had forgotten to restore Percy after disabling it on a day where the tests were behaving strangely, so this was basically done in passing.

R/utils.R
rpkyle: @josegonzalez The `:::` syntax is useful for accessing functions in other packages which are not exported. Usually it's something you'd apply interactively, and not a technique to apply within R package code. 

What's particularly strange is that this was added to access a function within the same package, no idea why I or anyone else did this, but in any event the CRAN check is careful to disallow this behaviour for submitted packages (and with good reason).

Very much a 🙈 moment.

R/utils.R
rpkyle: @josegonzalez `validate_keys` is an internal function, and only called from two Dash methods:
- `interpolate_index`
- `index_string`

The change makes it possible to use the same function for both of those methods, which pass different arguments to `validate_keys` (one passes the page index as a string with interpolation keys included, e.g. `{%app_entry%}` while the other passes in a vector of strings, in which the names of the keys are included, e.g. `app_entry`.

Both are matched against a list of necessary keys, and a useful error message thrown if the required keys are not present.

tests/testthat/test-index.R
rpkyle: @josegonzalez Good question. We previously raised an error in the last release, but this method was deprecated in that version, and now removed in this one.

PR #216, opened 2020-07-15
Remove no IDs warning

HammadTheOne:
Based on the discussion in plotly/dash-enterprise-docs#457, it would make more sense for new and experienced users to avoid this warning message altogether, especially in the cases where we have apps that don't require layouts or have the layout written first. This PR removes that warning, and updates the Dash R version and changelog.

Closes plotly/dash-enterprise-docs#457.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.6.4] - 2020-07-14
+**Changed**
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.6.4] - 2020-07-14
+**Changed**
+- Removes warning message when an app is launched without any component ID's present in the layout (for more details, see [#457](https://github.com/plotly/dash-enterprise-docs/issues/457).
```
alexcjohnson, line 7: DED isn't a public repo, so links to it are useless for most users. Just provide a brief (1-sentence) summary of the rationale.

PR #220, opened 2020-07-30
Fix debug mode warning output display

strboul:
I very much appreciate that the errors are caught as warnings in debug mode so they don't crash the running application, but it was annoying that I couldn't see where the user-defined errors occurred.

There was a bug in the `indexFromLast` object returning `NA`.

<details>
  <summary>Test application</summary>
  
```r
library(dash)
library(dashHtmlComponents)
library(dashCoreComponents)

app <- Dash$new()

app$layout(
  htmlDiv(
    list(
      dccInput(id = "inputID", value = "initial value", type = "text"),
      htmlDiv(id = "outputID")
    )
  )
)

fun <- function() {
  stop("this is an error message")
}

app$callback(
  output = list(id = "outputID", property = "children"),
  params = list(
    input(id = "inputID", property = "value"),
    state(id = "inputID", property = "type")
  ), function(x, y) {
    fun()
    sprintf("You've entered: '%s' into a '%s' input control", x, y)
  }
)

app$run_server(debug = TRUE)
```
</details>

Please let me know if there is any problem with this approach. I didn't run the unit & integration tests.


rpkyle:
@strboul Many thanks for this PR; I think the proposed edits are a nice improvement to this segment of the code (and kudos to you for diving into the stack tracing/error handling functions, which are among the less-friendly parts of the codebase).

💃

PR #224, opened 2020-08-15
Add support for callback graph improvements and timing

### LINKED ISSUE #223: Support for callback timing and general graph improvements
We should lay the foundations for supporting the features added by https://github.com/plotly/dash/pull/1179 as a follow-up to https://github.com/plotly/dash/issues/1176, which will permit timing of callback execution and add the ability to reorganize existing nodes on the graph when the dev tools UI is opened (among other improvements):

image

The changes required are relatively trivial and could be added concurrently with the upcoming pattern matching callbacks PR for our next CRAN release.

rpkyle:
This PR proposes to introduce additional context within the dev tools callback graph, as outlined previously in plotly/dash#1179, specifically

- a new `callback_context.record_timing` method is added, which permits capturing callback execution duration and adding it to `Server-Timing` headers (more info [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing)), as seen in figure below

Most of the changes are due to updates within `dash-renderer` based on [#1179](https://github.com/plotly/dash/pull/1179), including:
- `viz.js` is replaced with `cytoscape.js` via `react-cytoscape`; the callback graph is now interactive, and its layout/organization (e.g. when dragging/reordering nodes) persists throughout the session
- `CallbackGraphContainer` now subscribes to layout and paths so it can introspect the current state
- new reducers are added to compute profiling information and report all changed props (not just inputs)

Sample code (included within a callback), and a snapshot of the network pane illustrating callback profiling data within the Server Timing headers:

```r
    Sys.sleep(0.1)
    app$callback_context.record_timing('task_1', 0.1, 'The first task.')
    
    Sys.sleep(0.7)
    app$callback_context.record_timing('task_2', 0.7)
    
    Sys.sleep(0.2)
    app$callback_context.record_timing('task_3', 0.2, 'Cleanup task.')
```

image

To do:
- [x] restrict invocation of `callback_context.record_timing` to callbacks, as with `app$callback_context`

Closes #223.


R/dash.R
```diff
@@ -378,14 +378,15 @@ Dash <- R6::R6Class(
...
-                                        31536000) # 1 year
+            response$append_header('Cache-Control',
+                                   sprintf('public, max-age=%s',
+                                   31536000) # 1 year
             )
           } else if (!private$debug && !has_fingerprint) {
             modified <- as.character(as.integer(file.mtime(dep_path)))
 
-            response$set_header('ETag', modified)
+            response$append_header('ETag',
```
rpkyle, line 388: By using `append_header` instead of `set_header`, we can add the headers without removing others (see https://rdrr.io/cran/reqres/src/R/response.R).

R/dash.R
```diff
@@ -1056,6 +1087,41 @@ Dash <- R6::R6Class(
...
+            )
+          ))
+        })
+
+        self$server$on('request', function(server, request, ...) {
+          timing_information <- self$server$get_data('timing-information')
+          dash_total <- timing_information[['__dash_server']]
+          dash_total[['dur']] <- round(as.numeric(Sys.time() - dash_total[['dur']]) * 1000)
+          
+          request$response$append_header('Server-Timing',
```
rpkyle, line 1105: The `request` event isn't passed a `response` object here, because [it can be retrieved/modified via `request`](https://www.data-imaginist.com/2017/when-a-fire-starts-to-burn/):

> _In the same vein, the return values of request event handlers are ignored and the response is not passed to after-request event handlers as the response can be extracted directly from the request._

R/dash.R
```diff
@@ -1056,6 +1087,41 @@ Dash <- R6::R6Class(
...
+        self$server$on('before-request', function(server, ...) {
+          self$server$set_data("timing_information", list(
+            "__dash_server" = list(
+              "dur" = as.numeric(Sys.time()),
+              "desc" = NULL
+            )
+          ))
+        })
+
+        self$server$on('request', function(server, request, ...) {
```
rpkyle, line 1100: We will use `request` rather than `after-request`, which sounds desirable given our use of `after_request` in Dash for Python, because `after-request` in Fiery

> _Will trigger after the request event. This can be used to **inspect the response (but not modify it) before it is send to the client**. The handlers will receive the app itself as the server argument, the client id as the id argument, the request object as the request argument, and the response as the response argument. Any return value is discarded._

Compare with the behaviour of Fiery's `request`:

> _Will trigger after the before-request event. This is where the main request handling is done. **The return value of the last handler is send back to the client as response.** If no handler is registered a 404 error is returned automatically. If the return value is not a valid response, a 500 server error is returned instead. The handlers will receive the app itself as the server argument, the client id as the id argument, the request object as the request argument, and the list of values created by the before-event handlers as the arg_list argument._

R/dash.R
```diff
@@ -1162,10 +1228,19 @@ Dash <- R6::R6Class(
 
           # reset the timestamp so we're able to determine when the last cycle end occurred
           private$last_cycle <- as.integer(Sys.time())
+
+          # flush the context to prepare for the next request cycle
+          self$server$set_data("timing_information", list())
```
rpkyle, line 1233: [Unlike Flask's request context](https://flask.palletsprojects.com/en/1.1.x/reqcontext/), entries in the Fiery data store is not generally cleared at the end of the request cycle, so `self$server$set_data` appears to be persistent for the life of the session.

alexcjohnson: Feels to me like the `before-request` handler is enough to initialize this data, and if we want to clear it after the request we should just do that in the `request` handler immediately after using this info to set the `Server-Timing` headers.

rpkyle:
@Marc-Andre-Rivet I've completed my investigation, and as I discussed with @alexcjohnson, the `request` object is not easily modified within the handler (like `Dash`, it's an R6 class, and bindings can be locked). 

If you try to add a field on the object in place of what we were doing previously, i.e. `app$server$set_data("timing-information")`, you are likely to see this error:

```
Browse[1]> request$timing_information <- "test"
Error in request$timing_information <- "test" : 
  cannot add bindings to a locked environment
```

Similarly, unlike `Response`, there is no way to `append_header`:



... and you cannot force one in there:

```r
Browse[1]> request$headers$Server_Timing <- "test"
Error in (function ()  : 
  unused argument (base::quote(list(c("text/html", "application/xhtml+xml", "application/xml;q=0.9
```

An alternative I was considering is to hash the request object, then use the hash as a way to uniquely identify timing information within `self$server$get_data("timing-information")`, assuming the hash is reliably unique and not mutable. We already pull in `digest`, which has a function for computing the MD5 sum:

```r
digest::digest(request, "md5")
```

I did test twice, and upon initial request the hash was unique:

```r
Browse[1]> digest::digest(request, "md5")
[1] "d9da9c5617b829a5148351b27302f7fb"
```

```r
Browse[4]> digest::digest(request, "md5")
[1] "c1b00e7fdb85aeba08a1f6f32767453a"
```

I haven't verified that the hash is identical between `self$server$on('before-request', function(server, request, ...)` and `self$server$on('request', function(server, request, ...)`, but I'd guess that it should be.

rpkyle:
As I discussed with @alexcjohnson and @Marc-Andre-Rivet separately, I think we're going to revert to the original implementation which doesn't fuss around with evaluating functions in other environments (and setting attributes on the request object). I'll also write a test with three callbacks that are somewhat slow-running, to attempt to validate that the information recorded during a particular callback's execution is correctly matched within the server's data store.

The appropriate long-term solution is likely to submit a PR to provide an optional request context within `reqres` upstream, but that's out of scope for the current release cycle. We can reconsider this alternative at some point in the future.

rpkyle:
As discussed with @chriddyp, @Marc-Andre-Rivet, @alexcjohnson, I'll finish up with this PR and defer the integration test we discussed to the next DE release cycle. I'll open an issue to ensure it remains a priority and link back her.

R/dash.R
```diff
@@ -480,9 +481,9 @@ Dash <- R6::R6Class(
                                  file.size(asset_path))
         close(file_handle)
 
-        response$set_header('Cache-Control',
-                            sprintf('public, max-age=%s',
-                                    '31536000')
+        response$append_header('Cache-Control',
+                               sprintf('public, max-age=%s',
+                               '31536000')
```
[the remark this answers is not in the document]
rpkyle: :see_no_evil: Yikes, thanks for mentioning that. I'll try to be a bit more careful, even if R is tolerant about sloppy typing. Fixed in 7efe099

R/dash.R
```diff
@@ -831,9 +832,46 @@ Dash <- R6::R6Class(
...
+      timing_information <- self$server$get_data("timing-information")
+
+      if (name %in% timing_information) {
+        stop(paste0("Duplicate resource name ", name, " found."), call.=FALSE)
+      }
+
+      timing_information[[name]] <- list("dur" = round(duration * 1000), 
+                                         "desc" = description)
+
+      self$server$set_data("timing-information", timing_information)
```
alexcjohnson, line 872: This may be a dumb question from an R noob, but is this `set_data` necessary? Above you `get_data` into `timing_information`, then you add a new key to `timing_information`, but if it's still the same object, won't that new key already be reflected in the server data?

rpkyle: @alexcjohnson Assuming I understand your question, yes, it is necessary. R uses by pass-by-value semantics for the most part, so even if we update `timing_information`, that object still lives in the scope of `callback_context.record_timing`.

If we don't use `set_data` to store the updated object, the new information is lost when the function exits.

This might be interesting (though I haven't read it carefully myself):

https://homepage.divms.uiowa.edu/~luke/R/references.html

alexcjohnson:
@rpkyle this looks good! Agreed that we can defer testing the complex parallel callback timing / context test until later, but it would be nice to include a simple test that timing info makes its way to the response - just a straight port of https://github.com/plotly/dash/blob/dev/tests/integration/devtools/test_callback_timing.py, then I think we'll be good to go!

rpkyle:
> @rpkyle this looks good! Agreed that we can defer testing the complex parallel callback timing / context test until later, but it would be nice to include a simple test that timing info makes its way to the response - just a straight port of https://github.com/plotly/dash/blob/dev/tests/integration/devtools/test_callback_timing.py, then I think we'll be good to go!

Right, sorry for not including that here. I would not merge this without even a very basic test :grimacing: I'll handle that today. Thanks for taking a second pass!

rpkyle:
> @rpkyle this looks good! Agreed that we can defer testing the complex parallel callback timing / context test until later, but it would be nice to include a simple test that timing info makes its way to the response - just a straight port of https://github.com/plotly/dash/blob/dev/tests/integration/devtools/test_callback_timing.py, then I think we'll be good to go!

added in e0c0841

tests/integration/test_callback_instrumentation.py
```diff
@@ -0,0 +1,57 @@
...
+       app$callback_context.record_timing("pancakes", 1.23)
+       Sys.sleep(0.5)
+       return(y)
+     }
+ )
+
+ app$run_server(debug=TRUE)
+"""
+
+def test_rsci001_test_callback_instrumentation(dashr):
```
alexcjohnson, line 33: Non-blocking, but in principle this can be `dashr_server` rather than `dashr`, since you don't need the browser, just a server to make a request to.

alexcjohnson (approved):
Looks great! 💃

tests/integration/test_callback_instrumentation.py
rpkyle: Thanks for the heads-up, fixed in cc44b22.

Issue #131, opened 2019-09-24
Dash for R API does not currently allow user-defined routes


rpkyle:
As described in https://dash.plot.ly/integrating-dash and seen in [code examples](https://community.plot.ly/t/allowing-users-to-download-csv-on-click/5550/9) posted to our community forum, Dash for Python permits app developers to leverage Flask support for adding new routes. 

Dash for R currently lacks this functionality, so Dash for Python code like this:

```
@app.callback(Output('my-link', 'href'), [Input('my-dropdown', 'value')])
def update_link(value):
    return '/dash/urlToDownload?value={}'.format(value)

@app.server.route('/dash/urlToDownload') 
def download_csv():
    value = flask.request.args.get('value')
    # create a dynamic csv or file here using `StringIO` 
    # (instead of writing to the file system)
    strIO = StringIO.StringIO()
    strIO.write('You have selected {}'.format(value)) 
    strIO.seek(0)    
    return send_file(strIO,
                     mimetype='text/csv',
                     attachment_filename='downloadFile.csv',
                     as_attachment=True)
```

currently has no direct counterpart on the R side. We should expose the ability for application developers to define new routes within the Dash for R API. One possible option is through the use of `fiery` plugins. e.g. a `server_route` method could be added which exposes the ability to add plugins with new route handlers:

```
plugin <- list(
    on_attach = function(server) {
     router <- server$plugins$request_routr
     route <- Route$new()
     route$add_handler('all', '*', function(request, response, arg_list, ...) {
       message('Hello')
       TRUE
    })
    router$add_route(route, 1)
    },
  name = 'Hello_plugin',
  require = 'request_routr'
)
```

Thanks to @CanerIrfanoglu for raising this issue initially.

rpkyle:
A `server_route` method would greatly simplify redirects, as recently required by edits to the Dash for R documentation. Adding these is currently possible, but requires far more verbose and user-unfriendly code, e.g.

```
plugin <- list(
  on_attach = function(server) {
    router <- server$plugins$request_routr
    route <- routr::Route$new()
    redirect_getting_started <- function(request, response, keys, ...) {
      response$status <- 301L
      response$set_header('Location', '/getting-started')
      TRUE
    }
    redirect_getting_started_2 <- function(request, response, keys, ...) {
      response$status <- 301L
      response$set_header('Location', '/getting-started-part-2')
      TRUE
    }
    redirect_state <- function(request, response, keys, ...) {
      response$status <- 301L
[...]

PR #225, opened 2020-08-17
Add support for user-defined server routes

rpkyle:
This PR proposes to address #131, in order that Dash for R users may be able to create and modify routes as is currently possible in Dash for Python, thanks to Flask's customizable routing. Fiery supports routing "plugins", but it can be complicated to use them, or modify routes that already exist (see below). Two new methods are proposed in Dash for R:

- `server_route`: this method registers a new URL rule, as described by its `methods`, `url` and `handler` function
- `redirect`: this method can be used to redirect a path (single, or parameterized/with wildcards) to another path, as the handler is specified internally and returns a `301` (Moved Permanently) status within the response

Since Dash for R lacks the support for decorator methods that Python provides, the `handler` argument allows matching a handler function with a URL to be routed. In brief, the `server_route` method registers the URL routing information via `self$server$set_data("user-routes")`. Upon invoking `app$run_server`, Dash checks whether any `user-routes` are defined; if one or more routes exists, a plugin is generated on the fly, to insert all the routes in the order they were defined. Finally, the plugin is attached to the running server object. This process is repeated when the server shuts down/is aborted and restarted. The routing information remains intact, as it was stored via `set_data`.

The proposed changes are a compromise between making the code succinct and continuing to take advantage of `routr` features, while making it easier for users to add/remove routes and find documentation for this feature generally.

-----

## Previous interface

```r
... #code specifying packages, layout, callbacks omitted

plugin <- list(
  on_attach = function(server) {
    router <- server$plugins$request_routr
    route <- routr::Route$new()
    redirect_getting_started <- function(request, response, keys, ...) {
      response$status <- 301L
      response$set_header('Location', '/layout')
      TRUE
    }
    redirect_state <- function(request, response, keys, ...) {
      response$status <- 301L
      response$set_header('Location', '/basic-callbacks')
      TRUE
    }
    route$add_handler('get', '/getting-started', redirect_getting_started)
    route$add_handler('get', '/getting-started-part-2', redirect_state)
    route$add_handler('get', '/state', redirect_state)
    router$add_route(route, "redirects")
  },
  name = 'redirect_urls',
  require = 'request_routr'
)

app$server$attach(plugin)
app$run_server()
```

## Proposed interface

```r
... #code specifying packages, layout, callbacks omitted

app$server_route("/getting-started", function(request, response, keys, ...) {
  response$status <- 301L
  response$set_header('Location', '/layout')
  TRUE
})

app$server_route("/getting-started-part-2", function(request, response, keys, ...) {
  response$status <- 301L
  response$set_header('Location', '/basic-callbacks')
  TRUE
})

app$server_route("/state", function(request, response, keys, ...) {
  response$status <- 301L
  response$set_header('Location', '/basic-callbacks')
  TRUE
})

app$run_server()
```

The `redirect` method provides a simplification of the above:

```r
app$redirect("/getting-started", "/layout")
app$redirect("/getting-started-part-2", "/basic-callbacks")
app$redirect("/state", "/basic-callbacks")
```

To do:
- [x] tests to validate adding routes, resulting HTTP status codes and redirects (as applicable)
- [x] provide sufficient documentation and simplify syntax, ensure example is provided
- [x] update CHANGELOG.md
- [x] allow developers to pass in a function when using `app$redirect` in lieu of a string for the `new_path` argument, which would be evaluated as a string within `app$redirect`'s internal handler

### DIFF
--- a/R/dash.R
+++ b/R/dash.R
@@ -552,38 +552,57 @@ Dash <- R6::R6Class(
     },
 
     # ------------------------------------------------------------------------
-    # method to add custom server routes
+    # methods to add custom server routes
     # ------------------------------------------------------------------------
     #' @description
     #' Connect a URL to a custom server route
-    #' @details 
+    #' @details
     #' `fiery`, the underlying web service framework upon which Dash for R is based,
     #' supports custom routing through plugins. While convenient, the plugin API
     #' providing this functionality is different from that provided by Flask, as
     #' used by Dash for Python. This method wraps the pluggable routing of `routr`
     #' routes in a manner that should feel slightly more idiomatic to Dash users.
+    #' ## Querying User-Defined Routes:
+    #' It is possible to retrieve the list of user-defined routes by invoking the
+    #' `get_data` method. For example, if your Dash application object is `app`, use
+    #' `app$server$get_data("user-routes")`.
+    #'
+    #' If you wish to erase all user-defined routes without instantiating a new Dash
+    #' application object, one option is to clear the routes manually:
+    #' `app$server$set_data("user-routes", list())`.
     #' @param path Character. Represents a URL path comprised of strings, parameters
     #' (strings prefixed with :), and wildcards (*), separated by /. Wildcards can
     #' be used to match any path element, rather than restricting (as by default) to
     #' a single path element. For example, it is possible to catch requests to multiple
     #' subpaths using a wildcard. For more information, see \link{Route}.
-    #' @param handler Adds a handler function to the specified method and path.
+    #' @param handler Function. Adds a handler function to the specified method and path.
     #' For more information, see \link{Route}.
-    #' @param methods Character. A string indicating the request method (in lower case, 
-    #' e.g. 'get', 'put', etc.), as used by `reqres`. For more information, see \link{Route}.    
+    #' @param methods Character. A string indicating the request method (in lower case,
+    #' e.g. 'get', 'put', etc.), as used by `reqres`. The default is `get`.
+    #' For more information, see \link{Route}.
     #' @examples
     #' library(dash)
     #' app <- Dash$new()
     #'
-    #' # A handler to serve 418 errors; see https://tools.ietf.org/html/rfc7168
-    #' app$server_route("/teapot", function(request, response, keys, ...) {
-    #'   response$status <- 418L
-    #'   response$body('may be short and stout')
+    #' # A handler to redirect requests with `307` status code (temporary redirects); 
+    #' # for permanent redirects (`301`), see the `redirect` method described below
+    #' # 
+    #' # A simple single path-to-path redirect
+    #' app$server_route('/getting-started', function(request, response, keys, ...) {
+    #'   response$status <- 307L
+    #'   response$set_header('Location', '/layout')
+    #'   TRUE
+    #' })
+    #'
+    #' # Example of a parameterized redirect with a wildcard for subpaths
+    #' app$server_route('/getting-started/*', function(request, response, keys, ...) {
+    #'   response$status <- 307L
+    #'   response$set_header('Location', '/layout')
     #'   TRUE
     #' })
     server_route = function(path = NULL, handler = NULL, methods = "get") {
       if (is.null(path) || is.null(handler)) {
-        stop("The server_route method requires that a URL and handler function are specified. Please ensure these arguments are non-missing.", call.=FALSE)
+        stop("The server_route method requires that a path and handler function are specified. Please ensure these arguments are non-missing.", call.=FALSE)
       }
 
       user_routes <- self$server$get_data("user-routes")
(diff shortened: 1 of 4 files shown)


alexcjohnson:
This looks great - much nicer than the existing interface. I'm not sure about the utility of `show_routes` and `remove_routes` vs just documenting (or exposing some other way) `app$server$get_data("user-routes")` and letting users inspect or modify it directly.

Two other comments I already mentioned to @rpkyle on Slack:
- We may want to document this using a more concise anonymous function syntax:
```
app$server_route("/getting-started", function(request, response, keys, ...) {
  response$status <- 301L
  response$set_header('Location', '/layout')
  TRUE
})
```
- Redirects seem to be a primary use case here, so we may as well make them even easier:
```
app$redirect("/getting-started", "/layout")
```
I notice in dash-docs we also have some [parametrized redirects](https://github.com/plotly/dash-docs/blob/ce99aaa26ad0c2662386c99c5a51d5c9b5a0ba02/dash_docs/server.py#L107-L109) - in as far as these are simply replacing the head of the path and keeping the tail intact there may be some syntax we could use like:
```
app$redirect("/getting-started/*", "/layout/*")
```
For more complex situations I could imagine a solution where the output url is a function of the input path pieces, but that might get more complicated than it's worth - at that point perhaps we just document how to do this with `app$server_route`.

R/dash.R
```diff
@@ -550,6 +551,115 @@ Dash <- R6::R6Class(
...
+    #'   response$set_header('Location', '/layout')
+    #'   TRUE
+    #' })
+    #'
+    #' # Example of a parameterized redirect with a wildcard for subpaths
+    #' app$server_route('/getting-started/*', function(request, response, keys, ...) {
+    #'   response$status <- 307L
+    #'   response$set_header('Location', '/layout')
+    #'   TRUE
+    #' })
```
alexcjohnson, line 602: Is it obvious to users how you'd access the wildcard value, or even better a named parameter, from inside the handler? Might be nice if the example showed this.

tests/testthat/test-routes.R
```diff
@@ -0,0 +1,110 @@
+context("routes")
+
+test_that("URLs are properly redirected with app$redirect", {
+ library(dashHtmlComponents)
+
+ app <- Dash$new()
+
+ app$redirect("/foo", "/")
+ app$redirect("/bar/*", "/")
```
alexcjohnson, line 9: Likewise it would be nice if the wildcard was used in the redirect.

R/dash.R
rpkyle: hopefully fixed in 4b15b7a 😅

tests/testthat/test-routes.R
rpkyle: fixed in 1f5bd7a

rpkyle: @alexcjohnson If we're to use parameterized redirects in `app$redirect` as well, I imagine it will require some thought since `keys` exist only within the handler itself. In order for the app developer to just pass in the new path with a key we'd either
- parse the string for `:keys` and loop over them, constructing the new path inside of the handler passed to `app$redirect`
- allow developers to pass in a function (instead of a string), which we would then evaluate to a string in place of `new_path` within `app$redirect`'s internal handler

If it's best to handle this now, just let me know which option makes the most sense. I'd tend to favour the second one if possible, since parsing could get complicated in some cases. Alternatively, I could open an issue to address in a subsequent PR.

tests/testthat/test-routes.R
```diff
@@ -0,0 +1,110 @@
...
+ )
+
+ expect_equal(
+   response_bar$status,
+   301L
+  )
+
+ expect_equal(
+   response_bar$headers$Location,
+   "/*"
```
alexcjohnson, line 48: I thought the intent here was that `app$redirect("/bar/*", "/*")` would redirect `/bar/foo` to `/foo`?

tests/testthat/test-routes.R
alexcjohnson: I like the function option too. I'd do it now, I suspect we're going to need that form sooner rather than later in the docs.

rpkyle: @alexcjohnson OK, I've added this additional feature in 4d6c27d. It's somewhat complicated to implement but appears to work as intended. 

I've tried to streamline passing in a function, by ignoring the formals, since the handler *must* have a signature including `(request, response, keys, ...)` anyway, but I'll test with passing in extra arguments as `...` and see if I can get it to work. 

I'm not going to go overboard there, though; if it's not straightforward, passing arbitrary arguments won't be supported in this PR.

I'll also add a unit test for the feature. The current syntax looks like

```
app$redirect("/accounts/:user_id/*", function() paste0("/users/", keys$user_id))
```

tests/testthat/test-routes.R
rpkyle: Yeah, this was my own failure to understand the header specification; sorry about that. From reading the MDN link and the RFC linked below, I don't think wildcards are supported within `Location`. The test has been revised in 4a7f2f9.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location
https://tools.ietf.org/html/rfc2616#page-135

tests/testthat/test-routes.R
rpkyle: Documentation for this feature ☝️ added in f6fcbd1.

R/dash.R
```diff
@@ -550,6 +551,136 @@ Dash <- R6::R6Class(
...
+          response$status <- 301L
+          response$set_header('Location', new_path)
+          TRUE
+        }
+      }
+      user_routes[[old_path]] <- list("path" = old_path,
+                                      "handler" = handler,
+                                      "methods" = methods)
+
+      self$server$set_data("user-routes", user_routes)
```
alexcjohnson, line 681: Could possibly 🌴 a little more by calling `self$server_route` here?

rpkyle: @alexcjohnson Definitely, not sure why I didn't think of that but your suggestions have really helped to clean up the implementation. `app$redirect` is looking nice and lean now.

Issue #98, opened 2019-06-22
Add support for integrity and crossorigin attributes


rpkyle:
The `external_scripts` and `external_stylesheets` arguments for Dash for R currently lack support for attributes permitted in Dash for Python:
- `integrity` 
- `crossorigin`

These should be added for parity with the Python implementation of Dash. The HTML5 specification provides for a rich set of possible attributes for [`<link>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) and [`<script>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) tags:

`<link>` attributes:

- `as`
- `crossorigin`
- `disabled`
- `href`
- `hreflang`
- `importance`
- `integrity`
- `media`
- `referrerpolicy`
- `rel`
- `sizes`
- `title`
- `type`
- `methods`
- `prefetch`
- `target`

`<script>` attributes:

- `async`
- `crossorigin`
- `defer`
- `integrity`
- `nomodule`
- `nonce`
- `referrerpolicy`
- `src`
- `type`

Marc-Andre-Rivet:
Closed by #226

PR #226, opened 2020-08-26
Provide support for script and stylesheet attributes

rpkyle:
This PR proposes to address #98; a brief summary of changes:
- provides parity with the current Dash for Python interface, allowing either a list of strings or a list of lists to be passed -- the latter providing access to setting attributes for a given URL
- validates attributes for both `<script>` and `<link>` tags; if an invalid attributes are passed, Dash for R will present a helpful error message identifying which attributes should be omitted/corrected
- replaces `sprintf` with `glue`, which is slightly easier to read and as performant

Proposed syntax:

```r
app <- Dash$new(external_stylesheets = list(
                                         list(
                                           href="https://codepen.io/chriddyp/pen/bWLwgP.css",
                                           hreflang="en-us")
                                         )
)
```

...and a sample result:

image

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -542,28 +542,6 @@ generate_css_dist_html <- function(tagdata,
                                    prefix = NULL,
                                    as_is = FALSE) {
   attribs <- names(tagdata)
-  allowed_attribs <- c("as",
-                       "crossorigin",
-                       "disabled",
-                       "href",
-                       "hreflang",
-                       "importance",
-                       "integrity",
-                       "media",
-                       "referrerpolicy",
-                       "rel",
-                       "sizes",
-                       "title",
-                       "type",
-                       "methods",
-                       "prefetch",
-                       "target",
-                       "charset",
-                       "rev")
-  if (!all(attribs %in% allowed_attribs)) {
-      stop(sprintf("The following specified stylesheet attributes are invalid: ",
-                    paste0(setdiff(attribs, allowed_attribs), collapse=", ")), call. = FALSE)
-  }
   if (!(local)) {
     if (any(grepl("^(?:http(s)?:\\/\\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$",
         tagdata,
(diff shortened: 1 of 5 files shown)


R/utils.R
```diff
@@ -536,51 +536,98 @@ get_mimetype <- function(filename) {
...
+      href <- tagdata$src
+    else
+      href <- tagdata    
     href <- sub("^/", "", href)
     modified <- as.integer(file.mtime(local_path))
-    sprintf("<link href=\"%s%s?m=%s\" rel=\"stylesheet\">",
-            prefix,
-            href,
-            modified)
+    glue::glue('<link href="{prefix}{href}?m={modified}" rel="stylesheet">')
```
Marc-Andre-Rivet, line 586: Since at https://github.com/plotly/dashR/pull/226/files#diff-cbb7235afee10c3646161358a7655b0aR581 we handle the list case, I was expecting this to handle both the list and href-string cases like here https://github.com/plotly/dashR/pull/226/files#diff-cbb7235afee10c3646161358a7655b0aR571

R/utils.R
```diff
@@ -536,51 +536,98 @@ get_mimetype <- function(filename) {
...
+      href <- tagdata$src
+    else
+      href <- tagdata
     href <- sub("^/", "", href)
     modified <- as.integer(file.mtime(local_path))
-    sprintf("<script src=\"%s%s?m=%s\"></script>",
-            prefix,
-            href,
-            modified)
+    glue::glue('<script src="{prefix}{href}?m={modified}"></script>')
```
Marc-Andre-Rivet, line 630: Same question as https://github.com/plotly/dashR/pull/226/files#r478448814 -- `href <-` feels like 🍝 from `link`, use `src` instead for clarity.

rpkyle: fixed in b2ff7a1

R/utils.R
rpkyle: Excellent point, fixed in b2ff7a1.

rpkyle:
> This PR looks generally fine. Just one question that might be a misunderstanding on my part.
> 
> Missing:
> 
> * [x]  changelog
> * [x]  tests cases
> 
> Would expect test cases for the following scenarios:
> 
> * [x]  "string" href/src only
> * [x]  valid/invalid attributes w/ and w/o href local or remote
> 
> With expected behavior of app failing or page source / DOM introspection to check for appropriate attributes

fixed in 5f409ea aa63335 3f4296a

rpkyle:
@Marc-Andre-Rivet I think this one is more or less ready to go, but curious whether you feel it satisfies your earlier requests.

Marc-Andre-Rivet (approved):
💃 Looks good to me

PR #228, opened 2020-09-09
189 - Add Pattern Matching Callbacks for Dash R

### LINKED ISSUE #189: pattern matching callbacks support
i.e. https://github.com/plotly/dash/pull/1103/ but for R

HammadTheOne:
This PR adds front-end support in Dash for R for parsing and passing through Pattern Matching ID's and values into callback arguments, similar to plotly/dash#1103. It involves changes to the route handler for callbacks in Dash for R, callback context updates to correctly identify triggered component ID's and values, and dependency changes to allow input, output, and state id's to be passed as lists. 

An example of PMC for Dash R in action:
Peek 2020-09-08 13-39

This PR is still a work in progress, standard unit tests and callback validation have to be added before it is merged.

- [ ] Before merging, remove `wildcards_test.R` - Placeholder for easier local testing.

Closes #189


R/utils.R
```diff
@@ -890,29 +941,86 @@ removeHandlers <- function(fnList) {
...
+    inputs <- sapply(callback_elements$inputs, function(x) {
+      inputs_vector <- unlist(x)
+      setNames(list(inputs_vector[names(inputs_vector) == "value" | names(inputs_vector) == "value.index"]), paste(as.character(jsonlite::toJSON(x$id)), x$property, sep="."))
+    })
+  } else {
+    inputs <- sapply(callback_elements$inputs, function(x) {
+      inputs_vector <- unlist(x)
+      setNames(list(inputs_vector[names(inputs_vector) == "value" | names(inputs_vector) == "value.index"]), paste(as.character(jsonlite::toJSON(x[[1]]$id)), x[[1]]$property, sep="."))
+    })
+  }
```
rpkyle, line 1023: I suspect we could DRY this function up a bit as well. I was hoping to ask you about this block:

```r
if (startsWith(input_id, "{")){
  return(list(`prop_id` = x, `value` = value))
} else {
  return(list(`prop_id` = x, `value` = value))
}
```

Is it just me or did we end up with some 🍝 in one of those two branches? I kept staring at it thinking I must have missed something. 🤔

wildcards_test.R
```diff
@@ -0,0 +1,147 @@
+# Simple Example with ALL
```
rpkyle, line 1: Nice sample app! As far as tests go, I imagine we might want to consider
- unit tests in `tests/testthat` (for throwing errors when callbacks are misspecified, after `app$callback` is invoked, but before `app$run_server()` is used)
- a simple integration test to check functionality at adding/removing elements (like with Alex's to-do list app)
- at least some of the checks we implemented in https://github.com/plotly/dash/blob/master/tests/integration/devtools/test_callback_validation.py, we should probably ensure we error out on the same items in R as we do in Python (and maybe provide the same messages when possible)

rpkyle (changes requested):
Nice work! This feature required some serious work to get right, definitely not one of the simpler improvements to the framework in R, but definitely one of the most valuable.

The majority of my comments relate to DRYing up the code a bit, but there are some spots where the code is a little hard to follow without comments. In particular, we should probably review `setCallbackContext` since this function has gotten a lot more complex and I'm not completely sure what's happening in a few spots.

On the testing side of things, we'll definitely want to add a couple unit tests to ensure errors are thrown, and at least a basic integration test to confirm that a simple app using a pattern matching callback (like the to-do list example) works as expected. Really excited to see the arrival of this feature in R, one of the cooler improvements to Dash in quite a while.

R/utils.R
```diff
@@ -403,6 +403,57 @@ assert_valid_callbacks <- function(output, params, func) {
     stop(sprintf("The callback method requires that one or more properly formatted inputs are passed."), call. = FALSE)
   }
 
+  # Verify that 'input', 'state' and 'output' parameters only contain 'Wildcard' keywords if they are JSON formatted ids for pattern matching callbacks
+  valid_wildcard_inputs <- sapply(inputs, function(x) {
+    if (grepl("ALL|MATCH|ALLSMALLER", x$id)) {
```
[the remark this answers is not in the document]
rpkyle: (as we discussed offline, the concern here is that we might throw errors if a developer uses a valid name that happens to be a superstring of `ALL`, `MATCH`, or `ALLSMALLER`; it's a bit contrived, but easily avoided, so might as well address it either by ensuring `inputs` are screened before validation or via exact matching on the reserved words only -- as mentioned above)

R/utils.R
rpkyle: > I suspect we could DRY this function up a bit as well. I was hoping to ask you about this block:

As discussed offline, we're going to see if it's possible to DRY up the first part of the function, and see if there's a cleaner way to handle the pattern matching callbacks case within `setCallbackContext`:

```r
  } else if (length(callback_elements$inputs[[1]]) > 1) {
    inputs <- sapply(callback_elements$inputs, function(x) {
      inputs_vector <- unlist(x)
      setNames(list(inputs_vector[names(inputs_vector) == "value" | names(inputs_vector) == "value.index"]), paste(as.character(jsonlite::toJSON(x$id)), x$property, sep="."))
    })
  } else {
    inputs <- sapply(callback_elements$inputs, function(x) {
      inputs_vector <- unlist(x)
      setNames(list(inputs_vector[names(inputs_vector) == "value" | names(inputs_vector) == "value.index"]), paste(as.character(jsonlite::toJSON(x[[1]]$id)), x[[1]]$property, sep="."))
    })
  }
```

R/utils.R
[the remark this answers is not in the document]
rpkyle: After talking things over with @alexcjohnson, it became evident that I glossed over an important distinction relative to Dash for Julia and Python here. (Sorry about that.) In both of those languages, `ALL`, `MATCH` and `ALLSMALLER` are symbols, rather than strings. In R, I think we could achieve the same effect by doing

```r
ALL <- as.symbol("ALL")
```
which is then convertible to a string (for serializing into JSON) via

```r
dash::to_JSON(as.character(ALL))
```

to yield

```
["ALL"]
```

I'll make the needed changes this morning and push to this branch, I'm just about done, about to test with your sample application.

.Rbuildignore
```diff
@@ -5,3 +5,5 @@ Makefile
 announcement\.Rmd
 todo\.txt
 .github
+tests/integration
+tests/circleci
```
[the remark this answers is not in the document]
HammadTheOne: Added in be2e509.

R/dependencies.R
```diff
@@ -41,6 +50,9 @@ state <- function(id, property) {
 
 dependency <- function(id = NULL, property = NULL) {
   if (is.null(id)) stop("Must specify an id", call. = FALSE)
+  if (is(id, 'list')) {
+    id = sprintf('{"index":["%s"],"type":"%s"}', id[['index']], id[['type']])
```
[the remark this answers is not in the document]
HammadTheOne: Fixed in 7ea4f56.

R/utils.R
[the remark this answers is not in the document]
HammadTheOne: As we talked about earlier, I added your changes in their own commit and updated the callback validation logic to reflect the move to using `symbols` instead of character strings. Updated in 18192cd.

R/utils.R
```diff
@@ -403,6 +403,57 @@ assert_valid_callbacks <- function(output, params, func) {
...
+      } else {
+        return(TRUE)
+      }
+    })
+  }
+  
+  if (!all(valid_wildcard_output)) {
+    stop(sprintf("A callback output ID contains restricted pattern matching callback selectors ALL, MATCH or ALLSMALLER. 
+                 Please verify that it is formatted as a pattern matching callback list ID, or choose a different component ID."), call. = FALSE)
+  }
```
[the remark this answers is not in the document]
HammadTheOne: Updated in 504233a.

R/utils.R
```diff
@@ -403,6 +403,57 @@ assert_valid_callbacks <- function(output, params, func) {
     stop(sprintf("The callback method requires that one or more properly formatted inputs are passed."), call. = FALSE)
   }
 
+  # Verify that 'input', 'state' and 'output' parameters only contain 'Wildcard' keywords if they are JSON formatted ids for pattern matching callbacks
+  valid_wildcard_inputs <- sapply(inputs, function(x) {
```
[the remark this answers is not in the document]
HammadTheOne: Updated in 504233a.

R/dash.R
```diff
@@ -211,23 +211,37 @@ Dash <- R6::R6Class(
         #
         # https://cran.r-project.org/doc/FAQ/R-FAQ.html#Others:
         callback_args <- list()
-
+        
         for (input_element in request$body$inputs) {
-          if(is.null(input_element$value))
+          if ("id.index" %in% names(unlist(input_element))) {
+            unlisted_input <- unlist(input_element)
```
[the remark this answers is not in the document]
HammadTheOne: I'm a bit hesitant to change this part for a couple of reasons. For a normal callback, `request$body$inputs` is already a named list, and unlisting that would make handling regular callbacks more complicated as we then have to append a list of lists to `callback_args` for default callbacks. 

In the case of pattern matching callbacks, each `input_element` is a list of lists. By unlisting it we can access each of the wildcard inputs. I tried to use a vector to subset as you suggested, but while it works for some scenarios, on the initial call of the PMC we run into a subscript out of bounds error, and it might overcomplicate it to account for another condition. 

I'll try to figure out a way to simplify it and improve readability, but I think unlisting with a `for loop` in this case would have to be a necessary evil.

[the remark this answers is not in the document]
rpkyle: OK, trust your judgment (since I do as well). Though one remark I'd make is that the out of bounds error may be related to the current default values for callback inputs, which is `list(NULL)`. In that case, we could probably just ensure these values are handled properly, but don't worry about this for now.

We can always revisit the implementation later if we have an epiphany about making it a little more concise, for the moment it looks OK as-is.

tests/integration/callbacks/test_pattern_matching.py
```diff
@@ -0,0 +1,167 @@
...
+    return(htmlDiv(
+      lapply(1:length(test), function(x){
+        return(htmlDiv(sprintf("Dropdown %s = %s", x, test[[x]])))
+      })
+    ))
+  }
+)
+
+
+app$run_server(debug=T)
```
rpkyle, line 59:
```suggestion
app$run_server(debug=TRUE)
```

tests/integration/callbacks/test_pattern_matching.py
```diff
@@ -0,0 +1,167 @@
...
+"""
+
+allsmaller_app = """
+
+library(dash)
+library(dashCoreComponents)
+library(dashHtmlComponents)
+
+
+df <- read.csv('https://raw.githubusercontent.com/plotly/datasets/master/gapminder2007.csv', stringsAsFactors = F)
```
rpkyle, line 69: I always try to use `TRUE` and `FALSE` instead of `T` and `F`, even though the latter will work. Just a suggestion.

```suggestion
df <- read.csv('https://raw.githubusercontent.com/plotly/datasets/master/gapminder2007.csv', stringsAsFactors = FALSE)
```

tests/integration/callbacks/test_pattern_matching.py
```diff
@@ -0,0 +1,167 @@
...
+      htmlDiv(id = list("index" = n_clicks, "type" = "output-ex3"), children = list(unique(df$country)[n_clicks + 1]))
+    ))
+    
+    existing_children <- c(existing_children, list(new_children))
+  }
+)
+
+
+app$callback(
+  output(id = list("type" = "output-ex3", "index" = dash:::MATCH), property = "children"),
```
rpkyle, line 104: I was surprised to see `dash:::MATCH` since we explicitly exported these symbols in the `.Rd`, i.e. there are export lines:

https://github.com/plotly/dashR/blob/2d84a88d6182b316bea8a0afb7eadc4d566e7d2e/R/dependencies.R#L82-L92

There's an extra step I should have mentioned before, to trigger a `roxygen2` update to the `NAMESPACE` file:

```r
Rscript -e 'devtools::document()'
```

I went ahead and executed this step, then committed the updated documentation, so now the objects are exported and we can do instead:

```suggestion
  output(id = list("type" = "output-ex3", "index" = MATCH), property = "children"),
```

It turns out that CRAN won't allow the `:::` syntax in submitted packages, but no sweat, we're all set there now.

tests/integration/callbacks/test_pattern_matching.py
```diff
@@ -0,0 +1,167 @@
...
+  params = list(
+    input(id = list("type" = "filter-dropdown-ex3", "index" = dash:::MATCH), property = "value"),
+    input(id = list("type" = "filter-dropdown-ex3", "index" = dash:::ALLSMALLER), property = "value")
```
rpkyle, line 107:
```suggestion
    input(id = list("type" = "filter-dropdown-ex3", "index" = MATCH), property = "value"),
    input(id = list("type" = "filter-dropdown-ex3", "index" = ALLSMALLER), property = "value")
```

tests/integration/callbacks/test_pattern_matching.py
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+      return(
+        htmlDiv(sprintf("%s is the life expectancy of %s, and %s.", avgLifeExp, paste(all_values[-length(all_values)], 
+        collapse = " , "), paste(all_values[length(all_values)])))
+      )
+    }
+  }
+)
+
+app$run_server(debug=T)
```
rpkyle, line 135:
```suggestion
app$run_server(debug=TRUE)
```

tests/testthat/test-wildcards.R
```diff
@@ -24,5 +24,35 @@ test_that("HTML `data-*` & `aria-* ` wildcards are passed along to layout approp
   expect_equal(x$props$`data-foo`, 1)
 })
 
+
+test_that("restricted pattern matching callback selectors must be formatted as a list", {
```
rpkyle, line 28: It's OK to add these here, just wanted to note that the "wildcards" tested earlier in this script are the HTML wildcards, e.g. `data-*`, `html-*`.

tests/testthat/test-wildcards.R
```diff
@@ -24,5 +24,35 @@ test_that("HTML `data-*` & `aria-* ` wildcards are passed along to layout approp
...
+    htmlButton("Add Filter", id="add-filter", n_clicks=0),
+    htmlDiv(id="dropdown-container", children=list()),
+    htmlDiv(id="dropdown-container-output")
+  )))
+  
+  expect_error(
+    app$callback(
+      output(id="dropdown-container-output", property="children"),
+      params = list(
+        input(id=list("index" = dash:::ALL, "type" = "filter-dropdown"), property= "value")
```
rpkyle, line 40:
```suggestion
        input(id=list("index" = ALL, "type" = "filter-dropdown"), property= "value")
```

tests/testthat/test-wildcards.R
```diff
@@ -24,5 +24,35 @@ test_that("HTML `data-*` & `aria-* ` wildcards are passed along to layout approp
...
+      ),
+      display_output <- function(test){
+        return(htmlDiv(
+          lapply(1:length(test), function(x){
```
rpkyle, line 44: I generally use `seq_along(test)` instead of `1:length(test)`, for a better discussion of the reasons why than what I would likely provide, you might want to skim this chapter: https://r4ds.had.co.nz/iteration.html.

tests/testthat/test-wildcards.R
```diff
@@ -24,5 +24,35 @@ test_that("HTML `data-*` & `aria-* ` wildcards are passed along to layout approp
...
+            return(htmlDiv(test))
+          })
+        ))
+      }
+    ),
+    "Error: A callback input ID contains restricted pattern matching callback selectors ALL, 
+    MATCH or ALLSMALLER. Please verify that it is formatted as a pattern matching callback 
+    list ID, or choose a different component ID.",
```
rpkyle, line 52: Related to earlier comment, wonder if we still need this test now that we're using symbols rather than character strings.

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+    children[[n_clicks + 1]] <- new_dropdown
+    return(children)
+  }
+)
+
+
+app$callback(
+  output(id="dropdown-container-output", property="children"),
+  params = list(
+    input(id=list("index" = dash:::ALL, "type" = "filter-dropdown"), property= "value")
```
rpkyle, line 41:
```suggestion
    input(id=list("index" = ALL, "type" = "filter-dropdown"), property= "value")
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+    return(htmlDiv(
+      lapply(1:length(test), function(x){
+        return(htmlDiv(sprintf("Dropdown %s = %s", x, test[[x]])))
+      })
+    ))
+  }
+)
+
+
+app$run_server(debug=T, showcase = TRUE)
```
rpkyle, line 55:
```suggestion
app$run_server(debug=TRUE, showcase = TRUE)
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+    return(children)
+  }
+)
+
+
+app$callback(
+  output(id = list("index" = dash:::MATCH, "type" = "dynamic-output"), property= "children"),
+  params = list(
+    input(id=list("index" = dash:::MATCH, "type" = "dynamic-dropdown"), property= "value"),
+    state(id=list("index" = dash:::MATCH, "type" = "dynamic-dropdown"), property= "id")
```
rpkyle, line 129:
```suggestion
  output(id = list("index" = MATCH, "type" = "dynamic-output"), property= "children"),
  params = list(
    input(id=list("index" = MATCH, "type" = "dynamic-dropdown"), property= "value"),
    state(id=list("index" = MATCH, "type" = "dynamic-dropdown"), property= "id")
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+  ),
+  display_output <- function(value, id){
+    # ctx <- app$callback_context()
+    # print(jsonlite::toJSON(ctx, pretty = TRUE))
```
rpkyle, line 133:
```suggestion
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+    print(id)
+    print(value)
+    return(htmlDiv(sprintf("Dropdown %s = %s", id$index, value)))
+  }
+)
+
+app$run_server(showcase = T, debug = T, port = 8070)
```
rpkyle, line 140:
```suggestion
app$run_server(showcase = TRUE, debug = TRUE)
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+
+# Example with ALLSMALLER 
+
+
+library(dash)
+library(dashCoreComponents)
+library(dashHtmlComponents)
+
+
+df <- read.csv('https://raw.githubusercontent.com/plotly/datasets/master/gapminder2007.csv', stringsAsFactors = F)
```
rpkyle, line 152:
```suggestion
df <- read.csv('https://raw.githubusercontent.com/plotly/datasets/master/gapminder2007.csv', stringsAsFactors = FALSE)
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+    existing_children <- c(existing_children, list(new_children))
+  }
+)
+
+
+app$callback(
+  output(id = list("type" = "output-ex3", "index" = dash:::MATCH), property = "children"),
+  params = list(
+    input(id = list("type" = "filter-dropdown-ex3", "index" = dash:::MATCH), property = "value"),
+    input(id = list("type" = "filter-dropdown-ex3", "index" = dash:::ALLSMALLER), property = "value")
```
rpkyle, line 190:
```suggestion
  output(id = list("type" = "output-ex3", "index" = MATCH), property = "children"),
  params = list(
    input(id = list("type" = "filter-dropdown-ex3", "index" = MATCH), property = "value"),
    input(id = list("type" = "filter-dropdown-ex3", "index" = ALLSMALLER), property = "value")
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+  ),
+  display_output <- function(matching_value, previous_values){
+    previous_values_in_reversed_order = rev(previous_values)
+    all_values = c(matching_value, previous_values_in_reversed_order)
+    all_values = unlist(all_values)
+    
+    dff = df[df$country %in% all_values,]
+    avgLifeExp = round(mean(dff$lifeExp), digits = 2)
```
rpkyle, line 198: Not needed here (no missing values), but (something to think about in the future) I've gotten in the habit of using `mean(x, na.rm=TRUE)` since there are often missing values and I'd rather have an estimate than an `NA`. :laughing:

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+      )
+    } else {
+      return(
+        htmlDiv(sprintf("%s is the life expectancy of %s, and %s.", avgLifeExp, paste(all_values[-length(all_values)], collapse = " , "), paste(all_values[length(all_values)])))
+      )
+    }
+  }
+)
+
+app$run_server(showcase = T, debug = T, port = 8070)
```
rpkyle, line 216:
```suggestion
app$run_server(showcase = TRUE, debug = TRUE)
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+    output("list-container", "children"),
+    output("new-item", "value")
+  ),
+  params = list(
+    input("add", "n_clicks"),
+    input("new-item", "n_submit"),
+    input("clear-done", "n_clicks"),
+    state("new-item", "value"),
+    state(list("index" = dash:::ALL, "type" = "all"), "children"),
+    state(list("index" = dash:::ALL, "type" = "done"), "value")
```
rpkyle, line 255:
```suggestion
    state(list("index" = ALL, "type" = "all"), "children"),
    state(list("index" = ALL, "type" = "done"), "value")
```

wildcards_test.R
```diff
@@ -0,0 +1,281 @@
...
+        labelStyle = list("display" = "inline")
+      ),
+      htmlDiv(new_item, id = list("index" = "add", "type" = "all"), style = style_todo)
+    ), style = list("clear" = "both")))
+    return(list(new_list, ""))
+  }
+)
+
+
+app$run_server(showcase = T, debug = F, port = 8040)
```
rpkyle, line 281:
```suggestion
app$run_server(showcase = TRUE)
```

rpkyle (changes requested):
This round should go quickly -- these are primarily minor edits. The main points I have:
- do we still need to warn on use of reserved keywords for selectors? my guess is that this shouldn't be necessary given the use of symbols but I may be missing something
- we should export the symbols into the `NAMESPACE` so we don't have to use `:::`. (I updated the docs & `NAMESPACE`, so the edits below should be OK now.)
- might want to double-check syntax for `app$run_server()`, some of the port numbers in example apps are unusual, probably better to omit altogether?

Once these are done, @alexcjohnson might want to make a final pass over the code since it looks like everything we need is here (functional PMCs using keywords-as-symbols, working tests, and documentation of the feature).

R/dash.R
```diff
@@ -637,12 +663,43 @@ Dash <- R6::R6Class(
     #' containing valid JavaScript, or a call to [clientsideFunction],
     #' including `namespace` and `function_name` arguments for a locally served
     #' JavaScript function.
+    #'
+    #'
+    #' For pattern-matching callbacks, the `id` field of a component is written
+    #' in JSON-like syntax which describes a dictionary object when serialized
+    #' for consumption by the Dash renderer. The fields include a `type` and
+    #' an `index`, which describe the targets of the callback.
```
alexcjohnson, line 671: Does this mean we're only allowing `type` and `index` as keys (and in fact always using both of those keys)? That's much more restrictive than the Python version, where the fields can be any strings (except the empty string), and the values can be strings, numbers, and booleans, and should be serialized as such, ie no converting numbers or booleans to strings when these ids are serialized, and the keys should be sorted consistently. The id:
```
list("name" = "Alex", "age" = 43, "glasses" = TRUE)
```
should be stringified as:
```
'{"age":43,"glasses":true,"name":"Alex"}'
```
I don't actually know if the sorting matters in the end, but it might and we do enforce this in Python. Definitely flexible keys are important. We haven't made a lot of such examples because it quickly becomes a performance drag in the renderer and we're still working on that, but in principle you can have multiple wildcards, for example an output `list("row"=MATCH, "col"=MATCH, "type"="cell")` could depend on inputs `list("row"=MATCH, "type"="row header")` and `list("col"=MATCH, "type"="col header")`

rpkyle:
For the one failing unit test, looks related to the test for the error below; which at this point might be unnecessary using symbols. I might just remove this particular check, since character string IDs are fine, we'd be worried about collisions with the symbols (which, in my opinion, are exceedingly unlikely). 

Few R users create their own symbols, and it's also comparatively uncommon to use all caps for object names. A developer could create an object named `ALL`, `MATCH`, or `ALLSMALLER`, but there's nothing we can do in that case anyway. Our own object will get overloaded in that situation, e.g.

```r
> library(dash)
> ALL
ALL
> ALL <- "foo"
> ALL
[1] "foo"
```

The relevant CI failure and error:

image

```
"Error: A callback input ID contains restricted pattern matching callback selectors ALL, 
MATCH or ALLSMALLER. Please verify that it is formatted as a pattern matching callback 
list ID, or choose a different component ID."
```

R/utils.R
```diff
@@ -290,6 +290,22 @@ assert_no_names <- function (x)
...
 
+assertValidWildcards <- function(dependency) {
+  if (is.symbol(dependency$id)) {
+    result <- (jsonlite::validate(as.character(dependency$id)) && grepl("{", dependency$id))
+  } else {
+    result <- TRUE
+  }
+  if (!result) {
+    dependencyType <- class(dependency)
+    stop(sprintf("A callback %s ID contains restricted pattern matching callback selectors ALL, MATCH or ALLSMALLER. Please verify that it is formatted as a pattern matching callback list ID, or choose a different component ID.",
```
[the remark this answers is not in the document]
HammadTheOne: No, I can remove those now, and replace them with a different validation.

tests/testthat/test-wildcards.R
HammadTheOne: Thanks for pointing it out! Changed all of those to `seq_along` in 577deb4.

tests/testthat/test-wildcards.R
HammadTheOne: Removed in 577deb4.

rpkyle:
@HammadTheOne In addition to Alex's comments, do we currently handle/store `outputs_list`, `inputs_list`, and `states_list` in Dash for R now? I only ask because of what I see [here](https://github.com/plotly/dash/blob/5e717f1bd793ecaffc1200d97749b141feff98d2/dash/_callback_context.py#L35-L97).

In Python, it looks like these are stored via Flask's application context; I believe in Fiery the analog is assigning these values via `self$server$set_data` and retrieving them via `self$server$get_data` as in plotly/dashR#224.

R/dash.R
alexcjohnson: Looks like the code has been generalized to no longer require the keys to be `index` and `type`, which is great! Can we now generalize the comment above (or perhaps just remove that sentence) and also modify at least one test to use other keys?

tests/integration/callbacks/test_pattern_matching.py
```diff
@@ -0,0 +1,384 @@
+from selenium.webdriver.support.select import Select
```
alexcjohnson, line 1: Is this import used?

HammadTheOne: No, thanks for catching that, removed in d777ae1.

R/dash.R
HammadTheOne: Added in 210d3c1

alexcjohnson (approved):
💃  Fantastic, that's all from me!

rpkyle:
:woman_dancing: :shipit: :rocket:

rpkyle (approved):

PR #237, opened 2020-10-29
Fix setCallbackContext for wildcard and ordinary inputs

### LINKED ISSUE #236: setCallbackContext error occurs in pattern-matching callbacks with one or more input IDs as strings
In translating a Python demo app for pattern-matching callbacks with the following signature

```r
app$callback(
  output(id = list("index" = MATCH, "type" = "dynamic-output"), property = "figure"),
  params = list(
      input(id = list("index" = MATCH, "type" = "dynamic-dropdown-x"), property = "value"),
      input(id = list("index" = MATCH, "type" = "dynamic-dropdown-y"), property = "value"),
      input(id = "country", property = "value")
```

I encountered the error:

```r
values must be length 1,
 but FUN(X[[1]]) result is length 2
```

I was able to trace this error to `setCallbackContext`, and the following statement:

https://github.com/plotly/dashR/blob/add8dd3d93b23c0feb966e39f711ed58be256373/R/utils.R#L1065

I suspect that the issue is that we convert the IDs to JSON-like syntax later in the function, and so the length of the IDs is not uniformly `1`:

```r
> input_id                                                                                                                                          
[1] "country"

> callback_elements$inputs                                                                                                                          
[[1]]
[[1]]$id
[[1]]$id$index
[1] 0

[[1]]$id$type
[1] "dynamic-dropdown-x"


[[1]]$property
[1] "value"

[[1]]$value
[1] "year"


[[2]]
[[2]]$id
[[2]]$id$index
[1] 0

[[2]]$id$type
[1] "dynamic-dropdown-y"


[[2]]$property
[1] "value"

[[2]]$value
[1] "gdpPercap"


[[3]]
[[3]]$id
[1] "country"

[[3]]$property
[1] "value"

[[3]]$value
[1] "Bulgaria"
```

One option is to unlist the IDs when matching on their value. Alternatively, we could consider whether the ordering of the if/else logic within `setCallbackContext` is ultimately what we want.

Whatever solution we identify, we should ensure that a heterogeneous mix of IDs that are lists and vectors (i.e. using selectors or not) is allowed when describing callbacks so that this error does not occur.

HammadTheOne:
This PR addresses #236 and applies a fix to `setCallbackContext` to generalize the `id_match` and `prop_match` variables for both vectors and lists.

Closes #236

### DIFF
--- a/tests/integration/callbacks/test_pattern_matching.py
+++ b/tests/integration/callbacks/test_pattern_matching.py
@@ -489,3 +489,4 @@ def test_rpmc005_pattern_matching_graphs(dashr):
     dashr.select_dcc_dropdown('#\\{\\"index\\"\\:1\\,\\"type\\"\\:\\"dynamic-dropdown-x\\"\\}', "year")
     dashr.select_dcc_dropdown('#\\{\\"index\\"\\:1\\,\\"type\\"\\:\\"dynamic-dropdown-y\\"\\}', "pop")
     dashr.percy_snapshot("r-pmc-graphs")
+    dashr.wait_for_element('#\\{\\"index\\"\\:1\\,\\"type\\"\\:\\"dynamic-output\\"\\}')
--- a/tests/integration/callbacks/test_pattern_matching.py
+++ b/tests/integration/callbacks/test_pattern_matching.py
@@ -353,7 +353,8 @@ app$layout(
           )
         )
       ),
-      htmlDiv(id = "container", children=list())
+      htmlDiv(id = "container", children=list()),
+      htmlDiv(id = "output-delay")
     )
   )
 )
@@ -372,7 +373,10 @@ create_figure <- function(df, column_x, column_y, country) {
 }
 
 app$callback(
-  output(id = "container", property = "children"),
+  output = list(
+    output(id = "container", property = "children"),
+    output(id = "output-delay", property = "children")
+  ),
   params = list(
     input(id = "add-chart", property = "n_clicks"),
     state(id = "country", property = "value"),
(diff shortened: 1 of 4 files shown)


tests/integration/callbacks/test_pattern_matching.py
```diff
@@ -370,3 +479,14 @@ def test_rpmc004_pattern_matching_todo(dashr):
     dashr.find_element("#add").click()
     dashr.find_element('#\\{\\"index\\"\\:1\\,\\"type\\"\\:\\"done\\"\\}').click()
     assert dashr.wait_for_text_to_equal("#totals", "1 of 1 items completed - 100%")
+
+
+def test_rpmc005_pattern_matching_graphs(dashr):
+    dashr.start_server(graphs_app)
+    dashr.select_dcc_dropdown("#country", "Cameroon")
```
rpkyle, line 486: I often have a statement like ` dashr.wait_for_text_to_equal` initially before selecting anything, and then I have an `htmlDiv` with some content that renders which I can watch -- just in case there's a slight delay, in my (very limited) experience this seems to help avoid flaky tests, and can also take the place of a `sleep` statement.

rpkyle (approved):
Once the test passes, looks :ok_hand: to me. Thanks! :woman_dancing:

tests/integration/callbacks/test_pattern_matching.py
HammadTheOne: That sounds good, I'm going to make a small update to it with those elements, then I'll merge it once the final tests pass.

PR #240, opened 2020-11-16
Favicon fix

HammadTheOne:
This PR fixes the favicon issue described in plotly/dash-enterprise-docs#636. It adds a default Dash favicon and updates the tag to be set when no favicon.ico file is provided in the `assets` directory, in parity with Dash for Python. It also removes a redundant codeblock that may have been left behind from an earlier PR.


CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 All notable changes to `dash` will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## [0.9.0] - 2020-10-31
+## [0.9.1] - 2020-11-16
+### Fixed
+- Fixed a `favicon` response console error issue by adding a default favicon for Dash apps. [#240](https://github.com/plotly/dashr/pull/240)
```
rpkyle, line 7: Just a suggestion to tighten up the description here (since I think this resolved a regression as well):

```suggestion
- A regression which prevented favicons from displaying properly has been resolved, and a default Dash favicon is now supplied when none is provided in the `assets` directory. [#240](https://github.com/plotly/dashr/pull/240)
```

R/dash.R
```diff
@@ -1893,15 +1897,15 @@ Dash <- R6::R6Class(
       if ("/favicon.ico" %in% names(private$asset_map$other)) {
         favicon <- sprintf("<link href=\"/_favicon.ico\" rel=\"icon\" type=\"image/x-icon\">")
       } else {
-        favicon <- ""
+        favicon_url <- sprintf('\"%s_favicon.ico\"', self$config$requests_pathname_prefix)
```
rpkyle, line 1900: Have we tested this in a deployed application? Might not hurt to do so before merging the PR, something simple and quick on `dash-playground` should do (since we use `requests_pathname_prefix` here).

R/dash.R
```diff
@@ -1961,24 +1965,6 @@ Dash <- R6::R6Class(
...
-      if (!is.null(private$custom_index)) {
-        string_index <- glue::glue(private$custom_index, .open = "{%", .close = "%}")
-
-        private$.index <- string_index
-      }
-
-      else if (length(private$template_index) == 1) {
-        private$.index <- private$template_index
-      }
-
```
rpkyle, line 1981: Nice catch -- this is all dead code from before the index customization PR was merged, eh?

rpkyle (approved):
@HammadTheOne Just two minor comments -- I'd test this with a deployed app before merging, but assuming that works (and an app with another, non-default favicon displays properly), I'd say we're :ok_hand:.

Is there a test on the Python side for this handler? I feel like it wouldn't hurt to add a really basic test to lock in the desired behaviour here... since this is what led to us missing this issue until @VeraZab's wizardry identified it for us.

Here's an example of a recent [test](https://github.com/plotly/dashR/blob/master/tests/testthat/test-attributes.R) I wrote that introspects the tags and ensures that they're present; there's probably a better way to do it but this is what I used :sweat_smile: 

https://github.com/plotly/dashR/blob/edbcc040ffabb7956eeb4774e922794c53c557ce/tests/testthat/test-attributes.R#L39-L56

R/dash.R
HammadTheOne: I ended up testing it on `dash-core-docs`, and although there was a bit of a snag with the requests, I made a slight change and it's working as intended now both locally and deployed on DE.

R/dash.R
HammadTheOne: I think it was a copy/paste error during the index customization PR, it's just the same codeblock repeated as the one a few lines above it. Safe to cut :scissors:

PR #250, opened 2021-05-25
use correct non-vectorized version of logical and

daattali:


R/dash.R
```diff
@@ -116,9 +116,9 @@ Dash <- R6::R6Class(
       # ------------------------------------------------------------
       router <- routr::RouteStack$new()
       server$set_data("user-routes", list()) # placeholder for custom routes
- 
+
       # ensure that assets_folder is neither NULL nor character(0)
-      if (!(is.null(private$assets_folder)) & length(private$assets_folder) != 0) {
+      if (!(is.null(private$assets_folder)) && length(private$assets_folder) != 0) {
```
alexcjohnson, line 121: Does this change have any observable consequences (which we should lock down with a test), or is it just fixing poor practice that nonetheless happens to work?

daattali: As long as `private$assets_folder` is a single value rather than a vector/list, there should be no behavioural change. From my understanding, assets_folder should be a single value, but I don't see any checks for it so I'm not sure what currently happens when you pass in a vector

alexcjohnson: Yep, should be a single string. FWIW here's the docstring for the corresponding param in [Python](https://github.com/plotly/dash/blob/e727dd7e37d20147f6eefbdd35bd12319a026b27/dash/dash.py#L133-L138)

alexcjohnson (approved):
💃

alexcjohnson:
Ah shoot, didn't notice the base branch here is `master` rather than `dev`. Going to revert it so we don't have conflicts later on, but it's a good find that we should carry over to `dev` in a separate PR. @daattali please note there's a good deal of work in the `dev` branch that has yet to be released, as you're checking out the code. Most importantly #243

PR #262, opened 2021-06-21
Update `highlight.js` dependency for dash-table

HammadTheOne:
This PR updates the component libraries for Dash R and fixes the following issue with the `async-highlight.js` dependency sourcing when a `dash-table` is present within the layout:

image

This issue is caused by this shared dependency between `dash-core-components` and `dash-table`. When we merge the dependency artifacts into `inst/deps`, the newer version of the dependency sourced from `dash-table` is overwritten by the file sourced from `dash-core-components`. This PR adds a change so that only the version from `dash-table` is sourced for this dependency. As a longer-term solution we may want to sync up these dependencies between the two repos.


alexcjohnson:
After this change, does `dccMarkdown` work or does it yield a similar error? I'm worried that we may need to serve both files, ie put one of them in a different place or give it a different name, and adjust `.dashCoreComponents_js_metadata` or `.dashTable_js_metadata` accordingly.

HammadTheOne:
> After this change, does dccMarkdown work or does it yield a similar error? I'm worried that we may need to serve both files, ie put one of them in a different place or give it a different name, and adjust .dashCoreComponents_js_metadata or .dashTable_js_metadata accordingly.

I didn't notice any issues with `dccMarkdown` when using the different version of `highlight.js`. Testing out the examples in the docs, I didn't see any discernible difference between the two. I think if this has the possibility of happening with other dependencies then it might be worth exploring other ways to serve the files - we could even put dependencies into separate directories (eg. dcc, html, table) like we're doing with the Python monorepo, but that would require some larger changes.

alexcjohnson:
I don't understand the asymmetry - aside from a slight version difference in the underlying libraries ([10.3.1](https://github.com/plotly/dash-core-components/blob/72875fc254dcfe461d77ae649effb51a0ecb16c9/package.json#L51) vs [10.6.0](https://github.com/plotly/dash-table/blob/1e016de2dccda99bfefb100eb13a518af17af425/package.json#L83)) the two chunks look identical. The dash-table [lazy loader](https://github.com/plotly/dash-table/blob/1e016de2dccda99bfefb100eb13a518af17af425/src/dash-table/LazyLoader.ts#L8-L15) and [its target](https://github.com/plotly/dash-table/blob/dev/src/third-party/highlight.js) are exact  copies of the DCC [lazy loader](https://github.com/plotly/dash-core-components/blob/dev/src/utils/LazyLoader/hljs.js) and [its target](https://github.com/plotly/dash-core-components/blob/dev/src/third-party/highlight.js). So why would DCC work with the table chunk if the table doesn't work with the dcc chunk?

The DCC built chunk starts with:
```js
(window.webpackJsonpdash_core_components=window.webpackJsonpdash_core_components||[]).push(
```
whereas the table chunk starts with:
```js
(window.webpackJsonpdash_table=window.webpackJsonpdash_table||[]).push(
```

Which are pretty clearly incompatible with each other. So please check it out again - maybe your browser had a cached version of the DCC highlight chunk? I really think we're going to need to serve both, but we should be able to do so without restructuring the entire `deps` folder, just making a special case for this file.

You're right though that long-term we want to DRY this - see my comment https://github.com/plotly/dash-core/issues/323#issuecomment-867651335 for one possibility.

HammadTheOne:
You were right about the browser caching the `dcc` version of the highlight chunk. After clearing cache and relaunching the app, it fails to load a different chunk required by `dccMarkdown` (the component renders, but I'm guessing there will be problems with syntax highlighting). 

I think merging `DataTable` into `dcc` makes sense, but in the interim, I've changed the file names to `core-async-highlight.js` and `table-async.highlight.js` respectively (along with the sourcemaps), updated the JS metadata, and confirmed that it's working without any other console errors.

alexcjohnson:
> I've changed the file names to core-async-highlight.js and table-async.highlight.js respectively (along with the sourcemaps), updated the JS metadata, and tested that it's working without any other console errors.

@HammadTheOne are those changes ready to push? I think this PR will be ready at that point.

HammadTheOne:
@alexcjohnson Yep, sorry about that, got a little sidetracked. Changes have been pushed in 86105c7.

alexcjohnson (approved):
💃

PR #265, opened 2021-08-20
Add Dash 2 layout syntax wrappers and html tags

HammadTheOne:
This PR ports over the Dash2 changes and wrappers from https://github.com/daattali/dash2 into Dash for R, in a backwards compatible form. It also exports and enables the use of the pipe function from `magrittr`, along with the list of `html` tags and tag functions available for simplified layouts.

TODO:
- [x] Add basic unittests for layout and tags
- [x] Add Percy snapshot test
- [x] Update changelog

In a future PR, once this is more intertwined with the base Dash R code, we may want to add validation for the updated syntax and some add to the checks for the layout functions.


HammadTheOne:
@alexcjohnson  Any idea on why the CI checks aren't being reported to the Github status? The build passes on CircleCI: https://app.circleci.com/pipelines/github/plotly/dashR/848/workflows/453fedcf-5069-4c27-9118-8f0c69b4ec8a/jobs/905

And if I go to the `Checks` tab above, it's infinitely loading.

HammadTheOne:
Looks like the checks are passing now - I changed the `ci:circleci/test`  job check to the `build` workflow check for this branch to match the master one. 

Not sure if this was a recent change with CircleCI, but statuses are passed at a workflow level rather than a job level. For other repo's however, we are using job-level status checks (eg. plotly/dash) and those are being reported so it seems kind of contradictory. 

image

R/dash_2.R
```diff
@@ -0,0 +1,266 @@
...
+
+#' Create a Dash application
+#'
+#' This is a convenience function that returns a [`dash::Dash`] R6 object.
+#' For advanced usage, you can use the object as an R6 object directly instead
+#' of the functions provided by the `{dash2}` package.
+#'
+#' @param title _(character)_ The browser window title.
+#' @param update_title _(character)_ The browser window title while a callback
+#' is being processed. Set to `NULL` or `”"` if you don't want Dash to
```
alexcjohnson, line 11:
```suggestion
#' is being processed. Set to `NULL` or `""` if you don't want Dash to
```

R/dash_2.R
```diff
@@ -0,0 +1,266 @@
+# Dash 2 Syntax Functions (adapted from https://github.com/daattali/dash2)
```
alexcjohnson, line 1: We should certainly reference @daattali's work here, but I don't think we should call this part Dash 2 as though it's a separate thing from the rest of Dash... regardless of whether we call this version 1.0 or go straight to 2.0 to match Python.

R/dependencies.R
```diff
@@ -307,3 +307,271 @@ ALLSMALLER <- as.symbol("ALLSMALLER")
 #' @rdname selectors
 #' @export
 MATCH <- as.symbol("MATCH")
+
+
+# Dash 2 Syntax Functions
```
alexcjohnson, line 312: Why are these present both in `dash_2.R` and here?

tests/integration/test_render.py
```diff
@@ -29,12 +29,62 @@ def click_redo(self):
...
+      return("Hello Dash2")
+    }
+    else {
+      return("Hello Dash")
+    }
+  }
+)
+
+app %>%
+  add_meta(list(name = 'description', content = 'My App')) %>%
```
alexcjohnson, line 78: check for this meta tag in the test?

daattali:
I just saw this PR now. There was one local change I had that is very relevant to this PR so I just pushed it https://github.com/daattali/dash2/commit/63967402085dc6dd34d1bb4d989a8fb4bdfc9769

It allows conditional layout like this:

```
num <- 5
dash_app() %>%
    set_layout(num, " is ", if (num %% 2 == 1) "not ", "even") %>%
    run_app()
```

(weird example, I know)

R/utils.R
```diff
@@ -1536,3 +1536,38 @@ validate_keys <- function(string, is_template) {
...
+
+#' Is the given object a Dash app?
+#' @param x Any object.
+is_dash_app <- function(x) {
+  inherits(x, "Dash")
+}
+
+assert_dash <- function(x) {
+  if (!is_dash_app(x)) {
+    stop("You must provide a Dash app object (created with `dash::Dash$new()` or `dash2::dash_app()`)", call. = FALSE)
```
alexcjohnson, line 1550: Still a bunch of references to `dash2` here and in `wrappers.R` (only the link to https://github.com/daattali/dash2 should stay)

R/wrappers.R
```diff
@@ -0,0 +1,268 @@
...
+
+#' Create a Dash application
+#'
+#' This is a convenience function that returns a [`dash::Dash`] R6 object.
+#' For advanced usage, you can use the object as an R6 object directly instead
+#' of the functions provided by the `{dash2}` package.
+#'
+#' @param title _(character)_ The browser window title.
+#' @param update_title _(character)_ The browser window title while a callback
+#' is being processed. Set to `NULL` or `”"` if you don't want Dash to
```
alexcjohnson, line 11:
```suggestion
#' is being processed. Set to `NULL` or `""` if you don't want Dash to
```
Again in the copy that stayed 😉

R/dependencies.R
HammadTheOne: Removed in 0446581.

R/dash_2.R
HammadTheOne: Fixed in 0446581. I'm of the opinion that this be version 1.0, followed by 2.0 once we fully integrate the rest of Dean's changes.

R/wrappers.R
HammadTheOne: Thanks :hocho:  in ebdbe37

R/utils.R
HammadTheOne: Fixed in ebdbe37

PR #272, opened 2021-10-03
Dash R 1.0 release

HammadTheOne:
This PR merges in the changes from `dev` to `master` and updates the package, including the DESCRIPTION and CHANGELOG, in preparation for the Dash R 1.0 release to Github and CRAN.

- [ ]  Update CHANGELOG with updates based on commit history.
- [ ] Run checks for CRAN submission and formatting.


HammadTheOne:
Hey @alexcjohnson @daattali

I've updated the Dash R package and made the last few adjustments based on the checks that should prepare it for the CRAN re-submission and release. 

If there's anything else that stands out or anything else we want to add prior to that feel free to let me know. 

@daattali it might be a good idea to maybe install and run it in your local environment with the manual test apps, just to check for any inconsistencies that might exist between our development envs and R libraries.

daattali:
- [ ] Don't namespace `dash::` inside the dash package - it's redundant and confusing (I saw a few such cases in `/tests/manual/` and in documentation, there might be others)
- [ ] I would prefer to combine all the HTML components (htmlDiv, htmlBr, etc) into one file, and similarly all the interactive components (dcc*) into another file. This way will dramatically reduc ethe number of source files clutter
- [ ] I suggest updating the DESCRIPTION to use testthat version >3.0.0 , and then removing all `context()` calls from the test files (it's deprecated)
- [ ] The `tests/circleci/fixup_metadata.R` file should either be removed or moved to an appropriate folder
- [ ] In the `dashCoreComponents` package, the second argument of many components was `value`. Now the arguments are alphabetical and `value` has been moved to the last. This is a breaking change, and I strongly believe it should be fixed because having `value` as the second argument makes sense so that you don't have to always explicitly name it. @alexcjohnson @jackparmer this is now an issue with this package but this package is affected - where should I open an issue for this?
- [ ] The check that all states come after all inputs can be removed using the new flexible callback syntax https://github.com/plotly/dashR/blob/5c80285b90c24cf15375e6763b56425e9dc530ce/R/utils.R#L378-L381
- [ ] This is perhaps not directly related to this PR, but should be fixed: using `input()` instead of `output()` inside callbacks works. It shouldn't. There should be a check that anything supplied to the callback outputs is actually of class "output"

Let me know when the package is ready for re-testing

HammadTheOne:
> Don't namespace dash:: inside the dash package - it's redundant and confusing (I saw a few such cases in /tests/manual/ and in documentation, there might be others)

Fixed in d419517 and  3aa7abd.

> I would prefer to combine all the HTML components (htmlDiv, htmlBr, etc) into one file, and similarly all the interactive components (dcc*) into another file. This way will dramatically reduce the number of source files clutter

Concatenated components into package R scripts in cb93cdb. Note, this change requires the merge of https://github.com/plotly/dash/pull/1798 to make this a streamlined process, and also address the `NAMESPACE` generation with `devtools::document()`.

> I suggest updating the DESCRIPTION to use testthat version >3.0.0 , and then removing all context() calls from the test files (it's deprecated)

Updated in aff9b99.

> The tests/circleci/fixup_metadata.R file should either be removed or moved to an appropriate folder

Removed in 0ad3b24. 

> In the dashCoreComponents package, the second argument of many components was value. Now the arguments are alphabetical and value has been moved to the last. This is a breaking change, and I strongly believe it should be fixed because having value as the second argument makes sense so that you don't have to always explicitly name it. @alexcjohnson @jackparmer this is now an issue with this package but this package is affected - where should I open an issue for this?

I think the most appropriate repo for this issue would be in `dash`, with a label for `dashCoreComponents`. 

> The check that all states come after all inputs can be removed using the new flexible callback syntax

 Removed in 68427b3.

> This is perhaps not directly related to this PR, but should be fixed: using input() instead of output() inside callbacks works. It shouldn't. There should be a check that anything supplied to the callback outputs is actually of class "output"

Fixed with 68427b3.

daattali:
Awesome!

Do the CI checks run the unit tests, and that's why they failed previously? It looks like there's a build step in CircleCI , I assume that part of that includes running the tests, but I'm not able to see it since I don't have access to that account.

I think this PR is ready except for the breaking change of the order of parameters to dcc components changed. I think we should wait until we decide what to do about that, whether we'll fix it directly at the R level or higher up the chain

HammadTheOne:
@daattali I've reverted the prop order to the original order for the time being. We can talk about that more in detail next week, but if everything else looks good I think this should be ready to merge at this point.

PR #274, opened 2021-10-13
Remove html exports and update tags

HammadTheOne:
This PR removes direct html tag exports from the Dash R namespace, in favor of the `html$tag` list wrapper.

### DIFF
--- a/tests/integration/dopsa/test_dopsa.py
+++ b/tests/integration/dopsa/test_dopsa.py
@@ -8,7 +8,7 @@ app$layout(
     list(
       dccInput(id='input-1-state', type='text', value='Montreal'),
       dccInput(id='input-2-state', type='text', value='Canada'),
-      html$button(id='submit-button', n_clicks=0, children='Submit'),
+      html$button(id='submit-button', n_clicks=0, 'Submit'),
       html$div(id='output-state'),
       dccGraph(id='graph',
         figure=list(


daattali:
Old tests should still pass. You can use internal functions in testthat unit tests.
And old code should also still work , because old code would explicitly load `dashHtmlComponents`, in which case these functions would exist.

HammadTheOne:
@daattali I'm noticing one change in particular that I don't think is intended:

```
>s1 <- html$span("hmm", className = "value-output", `data-icon` = "fa-pencil")
>s2 <- html$span(children = list("hmm"), className = "value-output", `data-icon` = "fa-pencil")


>s1 
{
  "props": {
    "className": "value-output",
    "data-icon": "fa-pencil"
  },
  "type": "Span",
  "namespace": "dash_html_components",
  "propNames": ["children", "id", "n_clicks", "n_clicks_timestamp", "key", "role", "accessKey", "className", "contentEditable", "contextMenu", "dir", "draggable", "hidden", "lang", "spellCheck", "style", "tabIndex", "title", "loading_state", "data-icon"],
  "package": "dashHtmlComponents"
} 
> s2
{
  "props": {
    "children": "hmm",
    "className": "value-output",
    "data-icon": "fa-pencil"
  },
  "type": "Span",
  "namespace": "dash_html_components",
  "propNames": ["children", "id", "n_clicks", "n_clicks_timestamp", "key", "role", "accessKey", "className", "contentEditable", "contextMenu", "dir", "draggable", "hidden", "lang", "spellCheck", "style", "tabIndex", "title", "loading_state", "data-icon"],
  "package": "dashHtmlComponents"
} 
```

If the children property is explicitly set, it isn't passed to the `children` prop of the component object. However, if the string is passed without a name, it is passed to the dash component.

